PR #62, opened 2024-01-27
(external): [update] Catalog Products, fix categories:in schema

pvaladez:
API SPECS ISSUE

## What changed?

* For `GET` `/v3/catalog/products` endpoint, define `categories:in` query parameter schema as array of integers
* Schema currently defines an integer but it should be an array of integers.  For example, this is valid: `/v3/catalog/products?categories:in=24,25`
* See `id:in` schema: https://github.com/bigcommerce/docs/blob/47b47056ddb3274a7ab16c3db7b7c7749eacefd8/reference/catalog/products_catalog.v3.yml#L71-L78

## Release notes draft

* Allow multiple categories to be used with `categories:in` parameter on `/v3/catalog/products` endpoint.

## Anything else?

ping @slsriehl


reference/catalog/products_catalog.v3.yml
```diff
@@ -374,8 +374,12 @@ paths:
         - name: 'categories:in'
           in: query
           description: 'Filter items by categories. Use for products in multiple categories. For example, `categories:in=12`.'
```
slsriehl, line 376:
```suggestion
          description: 'Filter items by categories. Use for products in multiple categories. For example, `categories:in=12,15`.'
```

slsriehl (approved):

slsriehl:
Thanks @pvaladez!

PR #88, opened 2024-02-05
DEVDOCS-5687: [Update] changed option-id to an integer

bc-traciporter:
# [DEVDOCS-5687]

## What changed?
changed option id from a number to an integer

## Release notes draft

* 

## Anything else?

ping {names}

### DIFF
--- a/reference/carts.v3.yml
+++ b/reference/carts.v3.yml
@@ -2493,7 +2493,7 @@ components:
             type: object
             properties:
               option_id:
-                type: number
+                type: integer
               option_value:
                 type: integer
               name:


bc-shawnwang (approved):
thanks for the fix

bc-traciporter:
Thank you @bc-shawnwang. Do I need to add this change to the release notes?

bc-shawnwang:
@bc-traciporter I don't know how the release notes will express it so I'm not sure. Will adding it make them think it's a change of endpoint? (because it's not a change, we just fix the doc). Can release notes specify it? if so I think it's fine.

bc-traciporter:
@bc-shawnwang OK, sounds like we were just fixing a mistake. Is that correct?

bc-shawnwang:
@bc-traciporter yes, just fixing an doc mistake (or fix an inaccurate definition in the doc)

PR #132, opened 2024-02-28
CATALOG-8923: [update] Catalog, fix description in Create a variant using the product endpoint chapter

olysenko:
# CATALOG-8923: [update] Catalog, fix description in Create a variant using the product endpoint chapter

## What changed?
*  Fixed description about  option type that is set by default from radio buttons to rectangles in Create a variant using the product endpoint operation

## Release notes draft
*  Fixed description about  option type that is set by default from radio buttons to rectangles in Create a variant using the product endpoint operation

## Anything else?

@bc-traciporter


bc-traciporter (approved):
LGTM!

PR #138, opened 2024-02-28
DEVDOCS-5690: [Update] update add consignment to checkout endpoint

bc-traciporter:
# [DEVDOCS-5690]

## What changed?
Corrected the shipping consignment example

## Release notes draft

Bug Fix
Corrected the shipping consignment example
https://developer.bigcommerce.com/docs/rest-management/checkouts/checkout-consignments#add-consignment-to-checkout

* 

## Anything else?

ping {names}

### DIFF
--- a/reference/checkouts.v3.yml
+++ b/reference/checkouts.v3.yml
@@ -3983,9 +3983,9 @@ paths:
                       custom_fields:
                         - field_id: "string"
                           field_value: "string"
-                      line_items:
-                          item_id: "118f8774-387c-485e-a095-6ef76d5f1bbd"
-                          quantity: 1 
+                  - line_items:
+                      item_id: "118f8774-387c-485e-a095-6ef76d5f1bbd"
+                      quantity: 1 
               Pickup Consignment:
                 value: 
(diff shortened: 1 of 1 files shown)


reference/checkouts.v3.yml
```diff
@@ -3983,9 +3983,9 @@ paths:
                       custom_fields:
                         - field_id: "string"
                           field_value: "string"
-                      line_items:
-                          item_id: "118f8774-387c-485e-a095-6ef76d5f1bbd"
-                          quantity: 1 
+                  - line_items:
```
bc-andreadao, line 3986:
```suggestion
                    line_items: 
```

bc-andreadao (changes requested):
@bc-traciporter When I try out your changes in the spec tester, I get the following screenshot, which is not the same as what's posted on the ticket 
checkout consignment

I made a suggestion, which would give you the following:

checkout consignment 2

bc-traciporter:
Yes, @bc-andreadao you are correct. I confirmed this as well. Thank you.

PR #142, opened 2024-02-29
DEVDOCS-5810 [update]: Tax Provider API, update response body for adjust operation

bc-andreadao:
# [DEVDOCS-5810]

## What changed?
* Update response body schema for adjust operation
* Remove incorrect response example for adjust operation

[DEVDOCS-5810]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/tax_provider.yml
```diff
@@ -614,113 +614,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/request-adjust'
-              examples:
```
bc-andreadao, line 618: I removed the incorrect example, but if there's a good example we could use, we could add it in

theromulans: It’d be the same example as any of the others, although we could get fancy with it I suppose. I think it’s probably fine to simplify.

theromulans (dismissed):

bc-andreadao:
Thanks @theromulans

bc-traciporter (approved):
LGTM!

Issue #144, opened 2024-03-01
Orders V2 Create/Update does not accept values for tax_provider_id


AndrewBarber:
As highlighted on [BigCommerceDevs Slack](https://bigcommerce-devs.slack.com/archives/C012R9319AT/p1709311427053969), and confirmed by @bc-hnbarr on [forum here](https://support.bigcommerce.com/s/question/0D54O00007FNy8uSAD/tax-provider-id-via-order-api-post?language=en_US), OrdersV2 Create/Update does not accept a value for `tax_provider_id`.

The value should be shown as 'Read Only', and the enum reference removed to show that the value is not adhered to. 

(Although - strangely, supplying this does not cause the request to error, it is simply ignored.)

bc-tgomez:
Thanks @AndrewBarber for the catch and PR! 🙇  I'll get on this one.

bc-tgomez:
Your contribution should be live on our site shortly.

AndrewBarber:
Nice 👌 
I don't think I can self-close the issue. If you can that would be great! 👍

AndrewBarber:
Oh - I can! Cheers!

PR #146, opened 2024-03-01
#144 Add 'read-only' marker, Remove enum of accepted values for tax_p…

AndrewBarber:
…rovider_id on OrdersV2 create/update

#144 

## What changed?
* Add 'read-only' marker to tax_provider_id on OrdersV2 create/update
* Remove enum of accepted values for tax_provider_id on OrdersV2 create/update

## Release notes draft
* Ensures developers understand that any value will be ignored when using the endpoint.

## Anything else?
N/A

ping @bc-traciporter

### DIFF
--- a/reference/orders.v2.oas2.yml
+++ b/reference/orders.v2.oas2.yml
@@ -3850,15 +3850,12 @@ components:
           type: string
         tax_provider_id:
           description: |
+            Read-only.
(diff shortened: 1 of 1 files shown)


bc-tgomez (approved):
Wonderful! Thanks for the assist.

PR #152, opened 2024-03-05
DEVDOCS-5818 [update]: Catalog API, remove required flag from URL

bc-andreadao:
# [DEVDOCS-5818]

- Category Trees: remove required flag from URL


bc-traciporter (commented):
Hi @bc-andreadao maybe I am not understanding but I am thinking that you should remove url from the list of required variables.  
I see this in the CreateCategories schema and also in the Create Categories description section.

Creating a categories requires:
- name
- url
- tree_id or parent_id

bc-traciporter (approved):
LGTM

PR #177, opened 2024-03-19
[DEVDOCS-5413]: [revise] OrdersV2, is_deleted is a read-only value

bc-tgomez:
# [DEVDOCS-5413]

## What changed?

* Correct `is_deleted` definition to include that this is a read-only parameter.
*  Remove is_deleted from the body request for create an order.

## Release notes draft

* No PR needed for a read-only attribute fix

## Anything else?

ping {names}

[DEVDOCS-5413]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bc-traciporter:
@bc-tgomez Should you remove `is_deleted` or just change the description? If the request fails when `is_deleted` is included then maybe we need to remove it all together.
Also, while you are in the file or you can address it later, but the description for `variant_id` displays as ""
Screenshot 2024-03-19 at 1 23 16 PM

bc-traciporter (commented):
Left a comment

bc-tgomez:
You have a very good point. In the Jira ticket @bc-vincent-zhao mentioned adding the read-only attribute. I'm wondering if we can get Vincent to help make the decision. 👍  @bc-vincent-zhao should we remove this all together if it causes failure?

bc-vincent-zhao:
let's remove it 👍

reference/orders.v2.oas2.yml
```diff
@@ -4468,7 +4464,7 @@ components:
               example: '2001:db8:3333:4444:5555:6666:7777:8888'
               maxLength: 39
             is_deleted:
-              description: Indicates whether the order was deleted (archived). Set to to true, to archive an order.
+              description: This read-only parameter indicates if the order was deleted or archived.
```
bc-traciporter, line 4467: @bc-tgomez Can you check but I think you might need to delete this one?  In this case, the `is_deleted` field is inside order_Put which is used in a request body.

bc-traciporter (commented):
left a comment

Issue #173, opened 2024-03-16
Order Product API does not return order_product_id


lisastriker:
When i run a get request https://api.bigcommerce.com/stores/{store}/v2/orders/{order_number}/products. No order_product_id is found

This is my json result
`[
    {
        "id": 108,
        "order_id": 196,
        "product_id": 8250,
        "variant_id": 8215,
        "order_pickup_method_id": 0,
        "order_address_id": 97,
        "name": "Antenna, 3dBi Wifi emONE",
        "name_customer": "Antenna, 3dBi Wifi emONE",
        "name_merchant": "Antenna, 3dBi Wifi emONE",
        "sku": "ANT-008",
        "upc": "",
        "type": "physical",
        "base_price": "25.0000",
        "price_ex_tax": "25.0000",
        "price_inc_tax": "25.0000",
        "price_tax": "0.0000",
        "base_total": "1000.0000",
        "total_ex_tax": "1000.0000",
        "total_inc_tax": "1000.0000",
        "total_tax": "0.0000",
        "weight": "0.0800",
        "width": "0.0000",
        "height": "0.0000",
        "depth": "0.0000",
        "quantity": 40,
        "base_cost_price": "0.0000",
        "cost_price_inc_tax": "0.0000",
        "cost_price_ex_tax": "0.0000",
        "cost_price_tax": "0.0000",
        "is_refunded": false,
        "quantity_refunded": 0,
        "refund_amount": "0.0000",
        "return_id": 0,
        "wrapping_id": 0,
        "wrapping_name": "",
        "base_wrapping_cost": "0.0000",
        "wrapping_cost_ex_tax": "0.0000",
        "wrapping_cost_inc_tax": "0.0000",
        "wrapping_cost_tax": "0.0000",
        "wrapping_message": "",
        "quantity_shipped": 0,
        "event_name": null,
        "event_date": "",
        "fixed_shipping_cost": "0.0000",
        "ebay_item_id": "",
        "ebay_transaction_id": "",
        "option_set_id": null,
        "parent_order_product_id": null,
        "is_bundled_product": false,
        "bin_picking_number": "",
        "external_id": null,
        "fulfillment_source": "",
        "brand": "",
        "gift_certificate_id": null,
        "applied_discounts": [],
        "product_options": [],
        "configurable_fields": [],
        "discounted_total_inc_tax": "1000.0000"
    }
]`

[...]

PR #193, opened 2024-03-21
[GH-ISSUE] - [ #173] Order Product API does not return `order_product_id`

bc-tgomez:
# [DEVDOCS-5847]

## What changed?

* Add note indicating that the response `id` is required when creating order shipments

## Release notes draft

* Added a note to /v2/orders/{order_id}/shipments letting merchants know that the 'id' response is needed when creating order shipments.

## Anything else?

ping {names}

[DEVDOCS-5847]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/orders.v2.oas2.yml
```diff
@@ -472,7 +472,10 @@ paths:
       - $ref: '#/components/parameters/Accept'
       - $ref: '#/components/parameters/order_id_path'
     get:
-      description: Lists 50 order products on an order using `order_id`. By default, items sort from lowest to highest according to a newly created ID, separate from the `order_id` and the `product_id`.
+      description: |- 
+        Lists 50 order products on an order using `order_id`. By default, items sort from lowest to highest according to a newly created ID, separate from the `order_id` and the `product_id`.
```
bc-traciporter, line 476:
```suggestion
        Lists 50 order products on an order using `order_id`. By default, items are sorted from lowest to highest according to a newly created ID, separate from the `order_id` and the `product_id`.
```

bc-traciporter (approved):
I made one suggestion, you can take it or leave it as is.

PR #214, opened 2024-04-05
TAX-2050: Add performance/timeout section to tax provider API documentation

theromulans:
On the back of a recent incident we identified there was an opportunity to speak more about performance and our timeout expectations in our Tax Provider API documentation.

Here we are introducing a new section on this topic. Further comments inline.

### DIFF
--- a/docs/api-docs/provider-apis/tax/overview.mdx
+++ b/docs/api-docs/provider-apis/tax/overview.mdx
@@ -143,6 +143,16 @@ BigCommerce sends requests for tax estimates on product item, shipping, and hand
 
 [View the Estimate Taxes API reference](/docs/rest-contracts/tax#estimate-taxes).
 
+### Performance
+
+Merchants expect fast tax estimate responses to facilitate a streamlined checkout experience for their shoppers. Hence performance is an essential consideration for a tax provider.
+
+To protect the shopper checkout experience from overly slow tax responses, tax estimate requests will time out after a set period of time. The default timeout threshold of four seconds is intended to support a wide variety of use cases (e.g. large tax estimate requests).
+
+<Callout type="info">
+  You may request a lower timeout threshold to provide a better service to merchants.
+</Callout>
+
 ## Document submission
 
 Document submission enables you to persist a tax quote request, replace it with another one, or invalidate it if necessary.


docs/api-docs/provider-apis/tax/overview.mdx
```diff
@@ -143,6 +143,16 @@ BigCommerce sends requests for tax estimates on product item, shipping, and hand
...
 
+### Performance
+
+Merchants expect fast tax estimate responses to facilitate a streamlined checkout experience for their shoppers, hence performance is an essential consideration.
+
+To protect the shopper checkout experience from overly slow tax responses, tax estimate requests will timeout after a set period of time. The default threshold of four seconds is intended to support a wide variety of use cases (e.g. large tax estimate requests).
+
+<Callout type="info">
+  You may request a lower timeout threshold to provide a better service to merchants.
+</Callout>
```
theromulans, line 154: An important highlight is that tax providers can request a _lower_ timeout than the default four second timeout. 

We have intentionally not highlighted that they can (and have, in the past) request a _higher_ timeout (up to ten seconds maximum at this time) since we do want to discourage this.

docs/api-docs/provider-apis/tax/overview.mdx
```diff
@@ -143,6 +143,16 @@ BigCommerce sends requests for tax estimates on product item, shipping, and hand
 
 [View the Estimate Taxes API reference](/docs/rest-contracts/tax#estimate-taxes).
 
+### Performance
+
+Merchants expect fast tax estimate responses to facilitate a streamlined checkout experience for their shoppers, hence performance is an essential consideration for a tax provider.
+
+To protect the shopper checkout experience from overly slow tax responses, tax estimate requests will timeout after a set period of time. The default threshold of four seconds is intended to support a wide variety of use cases (e.g. large tax estimate requests).
```
theromulans, line 150: I've highlighted the _large cart_ scenario here since I think it's important to consider that not every tax request received will be equal - it may vary be merchant.

Indeed, this might be a legitimate reason to split out a tax provider's service into multiple components: one that services small carts, one that services large... and having appropriate timeouts set for each. That is, one size may not fit all.

That said, I think going into too much detail here could be excessive, and so I'm keen to keep it relatively light.

docs/api-docs/provider-apis/tax/overview.mdx
```diff
@@ -143,6 +143,16 @@ BigCommerce sends requests for tax estimates on product item, shipping, and hand
 
 [View the Estimate Taxes API reference](/docs/rest-contracts/tax#estimate-taxes).
 
+### Performance
+
+Merchants expect fast tax estimate responses to facilitate a streamlined checkout experience for their shoppers, hence performance is an essential consideration for a tax provider.
```
theromulans, line 148: One common issue we've explored with slow tax providers is whether they are conducting additional lookup operations _during a tax calculation_.

For example, actually calling out to BC for more information.

The most essential consideration there would be that _all the data needed to conduct a tax calculation should be in the tax request itself_. If a tax provider is calling out to BigCommerce for more data this is a signal they need to raise a feature request to include additional information in the request OR they need to source the data by other means (e.g. tax properties).

Obviously we can only guide so far and in so much detail - hence we are highlighting the core performance goal here only.

docs/api-docs/provider-apis/tax/overview.mdx
theromulans: My fear is that making this a field on provider registration will lead to more support chatter, specifically around _increasing_ this timeout value - although you could potentially apply the same thinking with even introducing this section.

I could see this as something we do once we achieve self-service tax provider registration (e.g. must be between 1 to 4 seconds - even if we've allowed for exceptions in some cases historically).

docs/api-docs/provider-apis/tax/overview.mdx
theromulans: We've allowed variations of 4-10 seconds in cases where we've highlighted exactly the above and the tax provider has opted to proceed with their ask. The most extreme case was the Midwest Goods tax provider (IGEN, from memory) that wanted to go even beyond 10 seconds, which we refused. That is, in their case reliable tax estimates (on large carts) were more important than speed.

So far these have been limited availability providers only, available only on a handful of stores - and so that seems acceptable to me, as long as the merchant knows what they are getting into and understands the consequences.

We've considered whether it's worth highlighting these timeout values - or performance metrics more generally - in the UI to give merchants better visibility into what they are getting into... with the intent of removing BigCommerce from the support conversation and directly encouraging tax providers to lift their performance. That is, taking more of an 'enablement' stance rather than getting in the way.

theromulans, line 150: > The default threshold of four seconds

While this is factual, and following on with some of @aleachjr's thinking, we have an opportunity here to reduce the 'default' timeout on new providers to a lower amount.

I could imagine a case for making this _two seconds_, for example, and removing the note about "You may request a lower timeout". This would put us in a stronger position going forward, and not inviting additional support chatter.

That is, because we're documenting this publicly we have a good opportunity to reconsider our position.

Another related consideration: if we want to explore how we can better service small carts vs large carts we might end up with a solution that somehow dynamically changes the timeout based on the number of items in the cart/quote request.

docs/api-docs/provider-apis/tax/overview.mdx
bc-andreadao, line 148:
```suggestion
Merchants expect fast tax estimate responses to facilitate a streamlined checkout experience for their shoppers. Hence performance is an essential consideration for a tax provider.
```

bc-andreadao: This is my only comment, otherwise looks good

docs/api-docs/provider-apis/tax/overview.mdx
theromulans: I'm going to run the numbers here on `AvalaraProvider` to see how they are actually tracking with regard to the general range of request timings.

theromulans: According to our recent metrics, our p99 for Avalara Provider is maxing-out around the 3 seconds mark, along with a couple of other providers (TaxCloud, Thomson Reuters). Meanwhile, the p99 mean is 1.15s for Avalara, though a substantial 2.26 for TaxCloud.

Vertex is also up there at 2.7s maximum (664ms mean) on p99.

p95 for our tracked platform providers, meanwhile, is mostly under 1 second (TaxCloud is currently an outlier at over 2, though a longer timeframe shows it mellow out also).

It's worth noting that these are to be regarded as higher grade tax providers, as opposed to the many limited availability providers we have on the platform - providers which generally have a far lower grade of support/technical implementation.

So, with that data up our sleeve and the context around the calibre of tax providers I think we can proceed with this documentation update as a starting point and address improvements as a separate initiative - in particular looking into what's going on with the p99 on our core providers (e.g. are these large carts?)

docs/api-docs/provider-apis/tax/overview.mdx
bc-andreadao, line 148:
```suggestion
Merchants expect fast tax estimate responses to facilitate a streamlined checkout experience for their shoppers. Hence performance is an essential consideration for a tax provider.
```
Fixes the run-on sentence

docs/api-docs/provider-apis/tax/overview.mdx
```diff
@@ -143,6 +143,16 @@ BigCommerce sends requests for tax estimates on product item, shipping, and hand
 
 [View the Estimate Taxes API reference](/docs/rest-contracts/tax#estimate-taxes).
 
+### Performance
+
+Merchants expect fast tax estimate responses to facilitate a streamlined checkout experience for their shoppers. Hence performance is an essential consideration for a tax provider.
+
+To protect the shopper checkout experience from overly slow tax responses, tax estimate requests will time out after a set period of time. The default timeout threshold of four seconds is intended to support a wide variety of use cases (e.g. large tax estimate requests).
```
theromulans, line 150: 👁️ Adjusted the wording here from `requests will timeout` to `requests will time out`, which I understand is more correct.

theromulans:
@bc-andreadao I think we're good to go when you are! 🙏 Thanks again.

docs/api-docs/provider-apis/tax/overview.mdx
theromulans: 🤔 Perhaps a refined stance above is:

_all the data needed to conduct a tax calculation should be in the tax request itself_

...becomes...

_all the BC data needed to conduct a tax calculation should be in the tax request itself_

bc-andreadao (approved):

PR #239, opened 2024-04-28
Devdocs 5916

Lucki2501:
# [DEVDOCS-5916]

## What changed?

* Replaced the incorrect "usd" `currency.code` in Carts API > changed to correct "USD"

## Release notes draft

* Merchants will be able to use the sample code for Carts with 'custom item' without running into an error.

## Anything else?

This is my first PR so hopefully all is well.


bc-traciporter (approved):
I confirmed your findings and you are correct.

PR #296, opened 2024-05-29
Fix response code for createShippingZones operation

funivan:
## What changed?
Change response code
image

## Release notes draft

Fixed a bug in the /v2/shipping/zones. Now the response code is 201 instead of 200

## Anything else?
-


theromulans (approved):
Thank you sir

slsriehl (approved):

PR #315, opened 2024-06-04
[DEVDOCS-5715]: [revise] Customers V3, Add origin_channel_id to the list of read only fields

bc-tgomez:
# [DEVDOCS-5715]

## What changed?

* Add `origin_channel_ids` to the list of read only fields

## Release notes draft

* Added origin_channel_id to the list of read-only fields in the Customers V3 API. This field is set upon customer creation and cannot be updated afterward.

## Anything else?

ping @

[DEVDOCS-5715]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/customers.v3.yml
+++ b/reference/customers.v3.yml
@@ -323,7 +323,7 @@ paths:
 
         * Attributes Values can not be updated using Update a Customer. Use the [Update customer attribute values](/docs/rest-management/customers/customer-attribute-values#upsert-customer-attribute-values) endpoint.
         * channel_ids -- Updating the list of channels a customer can access may create some side effects in a multi-storefront situation. This list determines which customer account we will use to authenticate a shopper given a channel.
-        * origin_channel_id -- Updating the string value doesn't return an error, nor does it update the value. This value is set when you create a customer.
+        * origin_channel_id -- This is an immutable value set as a reference to the channel of origin when a customer is created.
       summary: Update Customers
       tags:
         - Customers
--- a/reference/customers.v3.yml
+++ b/reference/customers.v3.yml
@@ -316,12 +316,14 @@ paths:
         * registration_ip_address
         * date_created
         * date_modified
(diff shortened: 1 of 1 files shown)


reference/customers.v3.yml
```diff
@@ -316,12 +316,14 @@ paths:
...
         * date_created
         * date_modified
+        * origin_channel_id
 
 
         **Notes**
 
         * Attributes Values can not be updated using Update a Customer. Use the [Update customer attribute values](/docs/rest-management/customers/customer-attribute-values#upsert-customer-attribute-values) endpoint.
         * channel_ids -- Updating the list of channels a customer can access may create some side effects in a multi-storefront situation. This list determines which customer account we will use to authenticate a shopper given a channel.
+        * origin_channel_id -- Updating the string value doesn't return an error, nor does it update the value. This value is set when you create a customer.
```
bc-tgomez, line 326: Hey @bc-akshatkaushik, do you think this needs to be added to the documentation? I've added `origin_channel_id` to the list of read-only values as part of the docs ticket submission. After testing it confirmed the statement.

bc-akshatkaushik:
```suggestion
        * origin_channel_id -- This is an immutable value set as a reference to the channel of origin when a customer is created.
```
I don't think `Updating the string value doesn't return an error, nor does it update the value` is useful to the user, as that is the expected behaviour for any xyz field you pass in to majority of the GRPC powered endpoints when that xyz field is not mapped to any field that the endpoint accepts

bc-akshatkaushik (approved):
Minor suggestion on phrasing but otherwise LGTM

PR #301, opened 2024-05-30
DEVDOCS-5911 [new]: MSF / Stencil, add channels object to themes

bc-andreadao:
# [DEVDOCS-5911]

## What changed?
- add channels to a theme's settings JSON object to `config.json` 

## Release notes draft

* The `settings` Stencil Theme object has a newly-released `channels` setting, which lets you add a storefront selector for shoppers to switch storefronts.

[DEVDOCS-5911]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #317, opened 2024-06-05
Revert "DEVDOCS-5911 [new]: MSF / Stencil, add channels object to themes"

bc-andreadao:
Reverts bigcommerce/docs#301


bc-traciporter (approved):
Approved

PR #322, opened 2024-06-07
DEVDOCS-5953: [update] add date created and date modified query parameters

bc-traciporter:
# [DEVDOCS-5953]

## What changed?
add date created and date modified query parameters

## Release notes draft

* 

## Anything else?

ping {names}

[DEVDOCS-5953]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #330, opened 2024-06-13
DEVDOCS-5953: [update] add date created and date modified

bc-traciporter:
# [DEVDOCS-5953]

## What changed?
Added parameters to the GET batch metafields endpoint.

## Release notes draft

* 

## Anything else?

ping {names}

[DEVDOCS-5953]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/locations.v3.yml
+++ b/reference/locations.v3.yml
@@ -1752,17 +1752,17 @@ components:
       schema:
         type: string
     date_modified_max:
-      name: 'date modified:max'
+      name: 'date_modified:max'
       description: |
-        'Query parameter that lets you filter by the maximum date_modified, for example, `2024-05-14T09:34:00`, `2024-05-14T09:34:00` or `2024-05-14`.' Returns metafields modified before this date.
+        'Query parameter that lets you filter by the maximum date modified, for example, `2024-05-14T09:34:00`, `2024-05-14T09:34:00` or `2024-05-14`. Returns metafields modified before this date.' 
       required: false
       in: query
       schema:
         type: string
     date_modified_min:
-      name: 'date modified:min'
+      name: 'date_modified:min'
       description: |
-        'Query parameter that lets you filter by the minimum date_modified, for example, `2024-05-14T09:34:00`, `2024-05-14T09:34:00` or `2024-05-14`.' Returns metafields modified after this date.
+        'Query parameter that lets you filter by the minimum date modified, for example, `2024-05-14T09:34:00`, `2024-05-14T09:34:00` or `2024-05-14`. Returns metafields modified after this date.' 
       required: false
       in: query
       schema:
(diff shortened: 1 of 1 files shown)


alex-gaiduchok:
@bc-rafsierk please take a look

reference/locations.v3.yml
```diff
@@ -1731,6 +1735,38 @@ components:
         enum:
           - asc
           - desc
+    date_created_min:
+      name: 'date_created:min'
+      description: |
+        'Query parameter that lets you filter by the minimum date created, for example, `2024-05-14T09:34:00Z`. Returns metafields created after this date.'
```
bc-rafsierk, line 1741: maybe `Query parameter that lets you filter by the date created which is before provided one`

FYI: it works like `>=`

reference/locations.v3.yml
```diff
@@ -1731,6 +1735,38 @@ components:
...
+      required: false
+      in: query
+      schema:
+        type: string
+    date_created_max:
+      name: 'date_created:max'
+      description: |
+        'Query parameter that lets you filter by the maximum date created, for example, `2024-05-14T09:34:00Z`. Returns metafields created before this date.'
```
bc-rafsierk, line 1749: As previously I would state that `Query parameter that lets you filter by the date created which is before provided one`
FYI: it works similarly to `<=`

reference/locations.v3.yml
bc-rafsierk, line 1741: Do we want to mention that we can provide just a date as well ?

bc-traciporter: We can, are you suggesting that we add a `date_created` parameter to the list? Would I need to do this just for the locations file or for the other APIs in [PR 322](https://github.com/bigcommerce/docs/pull/322/files).

bc-traciporter (commented):
@bc-rafsierk I left questions

reference/locations.v3.yml
bc-rafsierk: no, just if I would be a api user I would have no idea what `minimum date created` is. What I am suggesting is to say that this `min` means that the date provided in the filter will return metafields with the dates which are `after` the one used in filter. 
But if we usually just write minimum I am fine with that as well

bc-traciporter: Hi @bc-rafsierk , How does this sound? Is this clear?
Query parameter that lets you filter by the minimum date created. The date provided in the filter will return metafields after this date. Example format = 2024-05-14T09:34:00Z.

bc-rafsierk (approved):

bc-andreadao (approved):

PR #340, opened 2024-06-19
Remove duplicated `location_id` param

smcc-riccardoc:
## What changed?

* Remove duplicated parameter `location_id`

ping @slsriehl @bc-tgomez


bc-tgomez (approved):
Solid catch! Thank you!

PR #365, opened 2024-07-08
DEVDOCS-5923: [update] add customer access token

bc-traciporter:
# [DEVDOCS-5923]

## What changed?
Added customer access token to documentation

## Release notes draft

* 

## Anything else?

ping {names}

[DEVDOCS-5923]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CNanninga (changes requested):
In addition to in-line feedback:  I'm wondering what language we might want to add to the Customer Impersonation Token section to frame how we expect devs to treat CITs now.  I recommended moving the bullet points for why CATs are more secure to that section, so maybe that's sufficient.  But are we going to flat-out say that CITs are considered a "legacy" approach now, or something similar?  I'm not quie sure.  I'd want Nathan or someone similarly invested in the product strategy to weigh in.

bc-traciporter:
@jamie-mckenzie can you provide me with a customer access token OAS file so I can put it here with the other ones?
https://developer.bigcommerce.com/docs/rest-authentication/tokens

Screenshot 2024-07-18 at 3 01 26 PM

docs/start/authentication/graphql-storefront.mdx
```diff
@@ -66,17 +66,227 @@ content-type: application/json
 </Tab>
 </Tabs>
 
+#### Customer access tokens
+
+A customer access token is unique to an individual user's account because it represents an authenticated storefront session for GraphQL requests. You can obtain and use a customer access token only for server-to-server requests. Therefore, you must use the customer access token with a regular storefront token. A customer access token becomes invalid on all devices when you log out of a single device. 
+
+<Callout type="warning">
+Do not use this token for browser-side or client-side requests. Browser-side requests require using the [customer impersonation token](#customer-impersonation-tokens).
```
kskomro, line 74: for browser-side requests `SHOP_TOKEN` cookie is used, customer impersonation token is also for server-to-server requests.

docs/start/authentication/graphql-storefront.mdx
```diff
@@ -66,17 +66,227 @@ content-type: application/json
...
+    }
+}
+```
+</Tab>
+</Tabs>
+
+
+##### Using a Customer Access Token
+
+You can assign the customer access token obtained using one of the mutations described above to the `x-Bc-Customer-Access-Token` header for future queries. You can use this token for a specific shopper session access until it is invalidated. The token is invalidated when you use the logout mutation with the `x-Bc-Customer-Access-Token` header.
```
kskomro, line 206:
```suggestion
You can assign the customer access token obtained using one of the mutations described above to the `X-Bc-Customer-Access-Token` header for future queries. You can use this token for a specific shopper session access until it is invalidated. The token is invalidated when you use the logout mutation with the `X-Bc-Customer-Access-Token` header.
```
For headers it doesn't make a difference if letter is uppercase or lowercase, but I think usually in text we put this first `X` also big.

docs/start/authentication/graphql-storefront.mdx
```diff
@@ -66,17 +66,227 @@ content-type: application/json
...
+
+<Tabs items={['Request', 'Response']}>
+<Tab>
+
+```graphql filename="Example customer query" showLineNumbers copy
+POST https://{{storeDomain}}/graphql
+Authorization: Bearer {Storefront token}
+X-Bc-Customer-Id: {Customer ID}
```
kskomro, line 214: that should only be used with customer impersonation token, but for simple token only `X-Bc-Customer-Access-Token` header should be provided.

docs/storefront/graphql/index.mdx
```diff
@@ -591,7 +591,7 @@ Use normal GraphQL Storefront API tokens. According to the [Principle of least p
 
 ### I want to run requests from a server, and I need to support customer login
 
-Use a Customer Impersonation token and store it securely on your server like other secrets. When you need to run requests in the context of a particular customer (for example, if they've logged in to your application), send their BigCommerce Customer ID along with the request as the `x-bc-customer-id` header.
+Use a customer impersonation token and store it securely on your server like other secrets. When you need to run requests in the context of a particular customer (for example, if they've logged in to your application), send their BigCommerce Customer ID along with the request as the `x-bc-customer-id` header. You could also use a regular storefront token with a customer access token. You must assign the customer access token to `x-Bc-Customer-Access-Token` in the header.
```
kskomro, line 594: Customer impersonation token shouldn't be used for login I believe. Login mutation performed by this token will not return customerAccessToken, which I believe would be the whole reason for calling this mutation in server-to-server scenario.

CC: @zvuki @6juara9

docs/start/authentication/graphql-storefront.mdx
```diff
@@ -66,17 +66,226 @@ content-type: application/json
 </Tab>
 </Tabs>
 
+#### Customer access tokens
+
+A customer access token is unique to an individual user's account because it represents an authenticated storefront session for GraphQL requests. You can obtain and use a customer access token only for server-to-server requests. Therefore, you must use the customer access token with a regular storefront token. A customer access token becomes invalid on all devices when you log out of a single device. 
+
+<Callout type="warning">
+Do not use this token for browser-side or client-side requests. Browser-side requests require using the [customer impersonation token](#customer-impersonation-tokens) or the `SHOP_TOKEN` response cookie.
```
kskomro, line 74: customer impersonation token is not for browser-side requests! https://github.com/bigcommerce/docs/pull/365#discussion_r1686470808

also lower in the docs you have: 
``` The API will reject any customer impersonation token-authenticated requests that originate from a web browser and are not proxied. Use these tokens in NextJS applications; NextJS proxies requests by default. ```

Comment (bc-traciporter) [reply docs/start/authentication/graphql-storefront.mdx:74]: Thank you for catching this @kskomro . I removed the sentence.

File: docs/storefront/graphql/index.mdx (hunk 4 above)
Comment (bc-traciporter) [reply docs/storefront/graphql/index.mdx]: @kskomro Should I use the following for the answer?

Use a regular storefront token with a customer access token. You must assign the customer access token to `X-Bc-Customer-Access-Token` in the header to support customer login.

File: docs/start/authentication/graphql-storefront.mdx (hunk 5 above)
Comment (kskomro) [reply docs/start/authentication/graphql-storefront.mdx:74]: I think you removed the wrong one ;)
> The API will reject any customer impersonation token-authenticated requests that originate from a web browser and are not proxied. 
that was correct one and I used it to tell you that customer impersonation token is NOT allowed for browser-side requests.

Comment (kskomro) [reply docs/start/authentication/graphql-storefront.mdx:74]: > Browser-side requests require using the [customer impersonation token]

this is not true.

File: docs/start/authentication/graphql-storefront.mdx [hunk 6]
```diff
@@ -66,17 +66,226 @@ content-type: application/json
...
+##### Obtaining a customer access token
+
+There are two options to obtain a customer access token.
+
+1. Login mutation
+2. Login with a JWT
+
+###### Login mutation
+
+Enter your user email and password to use the login mutation. The login mutation will return a customer access token in response to login actions as part of the GraphQL body instead of a cookie header. From there, you can store the customer access token in the presentation layer's session management system and send it with future GraphQL requests.
```
bookernath, line 86: This behavior varies based on whether the request is made from a browser or not. If the login() mutation request is from a browser, we will not return the CAT in the body, and will instead set a cookie.

If the request does not come from a browser and instead is used in a server-to-server context, the token will be returned in the body.

docs/start/authentication/graphql-storefront.mdx
```diff
@@ -66,17 +66,226 @@ content-type: application/json
...
+    }
+}
+```
+</Tab>
+</Tabs>
+
+
+##### Using a Customer Access Token
+
+You can assign the customer access token obtained using one of the mutations described above to the `X-Bc-Customer-Access-Token` header for future queries. You can use this token for a specific shopper session access until it is invalidated. The token is invalidated when you use the logout mutation with the `X-Bc-Customer-Access-Token` header.
```
bookernath, line 206: _or_ if it expires due to time

docs/start/authentication/graphql-storefront.mdx
```diff
@@ -66,17 +66,226 @@ content-type: application/json
...
+
 ### Using auto-generated tokens on Stencil storefronts
 
 On Stencil storefronts, you can access a token at render time and pass the token to client-side code using the `{{settings.storefront_api.token}}` Handlebars property. This auto-generated token has an expiry period of 24-48 hours and will rotate before expiration.
 
+
 ## Customer impersonation tokens
 
-You can use customer impersonation tokens to authenticate requests to the GraphQL Storefront API in server-to-server and headless interactions. The API will reject any customer impersonation token-authenticated requests that originate from a web browser and are not proxied. Use these tokens in NextJS applications; NextJS proxies requests by default.
+<Callout type="warning">
```
bookernath, line 277: This section should have a callout that not all customer features will be supported when using a CIT in the future; features that are more security-sensitive will only work with CAT.

docs/start/authentication/graphql-storefront.mdx
```diff
@@ -66,17 +66,226 @@ content-type: application/json
...
 
-Customer impersonation token-authenticated requests can query store information for any customer by passing the customer ID using the `X-Bc-Customer-Id` header. Information that you request will be specific to the subject customer, including product pricing and availability, customer account details, and more.
+Customer impersonation token-authenticated requests will be specific to the customer, including product pricing and availability, customer account details, and more.
 
-It is not necessary to generate a new token for each customer ID. You may use a single token at any given time for your application, and specify a distinct customer ID for each set of requests.
+Generating a new token for each customer ID is optional. You may use a single token at any given time for your application and specify a distinct customer ID for each set of requests.
```
bookernath, line 288: I wouldn't even call it optional, there's no good reason to do it.

I would say "is not necessary".

docs/storefront/graphql/index.mdx
```diff
@@ -591,7 +591,7 @@ Use normal GraphQL Storefront API tokens. According to the [Principle of least p
 
 ### I want to run requests from a server, and I need to support customer login
 
-Use a Customer Impersonation token and store it securely on your server like other secrets. When you need to run requests in the context of a particular customer (for example, if they've logged in to your application), send their BigCommerce Customer ID along with the request as the `x-bc-customer-id` header.
+Use a customer impersonation token and store it securely on your server like other secrets. When you need to run requests in the context of a particular customer (for example, if they've logged in to your application), send their BigCommerce Customer ID along with the request as the `X-Bc-customer-id` header. You could also use a regular storefront token with a customer access token. You must assign the customer access token to `X-Bc-Customer-Access-Token` in the header.
```
bookernath, line 594: I would flip this so it sounds like CAT is the recommended option and CIT is another option, but not the recommendation.

docs/storefront/graphql/index.mdx
```diff
@@ -591,7 +591,7 @@ Use normal GraphQL Storefront API tokens. According to the [Principle of least p
 
 ### I want to run requests from a server, and I need to support customer login
 
-Use a Customer Impersonation token and store it securely on your server like other secrets. When you need to run requests in the context of a particular customer (for example, if they've logged in to your application), send their BigCommerce Customer ID along with the request as the `x-bc-customer-id` header.
+The recommended option is to use a regular storefront token with a customer access token. You will need to assign the customer access token to `X-Bc-Customer-Access-Token` in the header. The other possible but least preferred option is to use a customer impersonation token and store it securely on your server like other secrets. When you need to run requests in the context of a particular customer (for example, if they've logged in to your application), send their BigCommerce Customer ID along with the request as the `X-Bc-customer-id` header. This option provides less security but is available to use.
```
bookernath, line 594: 🍹 
You will need to assign the customer access token to `X-Bc-Customer-Access-Token` in the header.

-->

You can exchange customer credentials for a customer access token using the login() mutation and send the token on future requests using the `X-Bc-Customer-Access-Token` header.

bookernath (approved):

bookernath (approved):

PR #376, opened 2024-07-15
DEVDOCS-5558 [new]: Widget theme regions, add channel_id query parameter

bc-andreadao:
# [DEVDOCS-5558]

## What changed?

* Add  channel_id query parameter to Get Theme Regions endpoint

## Release notes draft

@bigcommerce/team-merchandising

### DIFF
--- a/reference/widgets.v3.yml
+++ b/reference/widgets.v3.yml
@@ -469,8 +469,16 @@ paths:
       parameters:
         - $ref: '#/components/parameters/Accept'
         - $ref: '#/components/parameters/RequiredTemplateFile'
+        - $ref: '#/components/parameters/ChannelID'
 components:
   parameters:
+    ChannelID: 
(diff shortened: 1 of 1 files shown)


reference/widgets.v3.yml
```diff
@@ -502,7 +510,7 @@ components:
         type: string
     FilterTemplateFileParam:
       in: query
-      name: template_file
+      name: template_file
```
bc-andreadao, line 513: From the ticket, it seems like this query parameter should also be added to the [Get Theme Regions](https://developer.bigcommerce.com/docs/rest-content/widgets/regions) endpoint? Currently, the docs have `templateFile` as the only query parameter

bc-erich (approved):

bc-traciporter (approved):
Approved

PR #395, opened 2024-07-24
Remove inaccurate note about deleting consignment unassociated a cart from a customer

CNanninga:
The API reference for deleting a Consignment contains a note that deleting a checkout's last consignment removes the cart from its assigned customer.  This has been tested and found to be inaccurate, and Valentin Dellangela verified it should be removed.


bc-traciporter (approved):
Approved

PR #399, opened 2024-07-25
Update description of TaxQuote::id

theromulans:
## What changed?
* Adjust description of the `TaxQuote::id` field to describe that we also accept external ID values.
* Also describe how this value is used with adjust and void requests.
* Bonus: adjust remaining outlier casing concerns for "Tax Provider" => "tax provider".

Based on tax partner feedback and a significant confusion internally.

## Release notes draft
* Clarify accepted Tax Provider API quote ID values and usages.

### DIFF
--- a/reference/tax_provider.yml
+++ b/reference/tax_provider.yml
@@ -1120,7 +1120,7 @@ components:
       properties:
         code:
           type: string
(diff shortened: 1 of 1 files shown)


bc-andreadao (approved):
Thanks @theromulans

PR #410, opened 2024-07-31
Change location of theme uuiid path param to be endpoint-wide & remove site_id:in query param

markcmurphy:
## What changed?
* `/themes/{uuid}/configurations` only has a single method, so you'd think it wouldn't matter whether you put the `uuid` path param to be method-specific or endpoint-wide. Still, the request runner expects it to be in the endpoint-scoped param object. 
* removed `site_id:in` because it is marked as required, yet not accepted by the endpoint (see screenshot below). Also, I could find no mention of it in the api-proxy-java repo. You might want to confirm with that team, though.

image


bc-tgomez (approved):

bc-tgomez:
Thanks @markcmurphy

PR #490, opened 2024-09-03
chore(shipping): SHIPPING-3163 Add documentation for checkout settings endpoint

payaalaee:
# [SHIPPING-3163]

## What changed?

- Added api docs for v3 shipping checkout settings (`shipping/settings` endpoint)

## Anything else?
API request and response with `Get` and `Post`:

https://github.com/user-attachments/assets/78f24e8f-cc40-458b-a911-a7fecb565b9b

ping @bigcommerce/team-shipping @bc-andreadao 

[SHIPPING-3163]: https://bigcommercecloud.atlassian.net/browse/SHIPPING-3163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/shipping.v3.yml
+++ b/reference/shipping.v3.yml
@@ -3,13 +3,14 @@ info:
   version: ''
   title: Shipping V3
   description: V3 REST API shipping endpoints.
-  termsOfService: 'http://www.bigcommerce.com/terms' 
-  contact: 
-    name: BigCommerce 
-    url: 'https://www.bigcommerce.com' 
-    email: support@bigcommerce.com 
+  termsOfService: 'http://www.bigcommerce.com/terms'
+  contact:
+    name: BigCommerce
+    url: 'https://www.bigcommerce.com'
+    email: support@bigcommerce.com
 tags:
   - name: Customs Information
+  - name: Shipping Settings
 security:
   - X-Auth-Token: []
 servers:
@@ -134,14 +135,66 @@ paths:
             examples:
               Example:
                 value:
-                - product_id: 77
-                  country_of_origin: US
-                  commodity_description: Baseball caps
-                  international_shipping: true
-                  hs_codes:
-                    CA: '508313'
-                    AU: '817355'
-                    ALL: '501000'
+                  - product_id: 77
+                    country_of_origin: US
+                    commodity_description: Baseball caps
+                    international_shipping: true
+                    hs_codes:
+                      CA: '508313'
+                      AU: '817355'
+                      ALL: '501000'
+
+  '/shipping/settings':
+    parameters:
+      - $ref: '#/components/parameters/Accept'
+
+    get:
+      operationId: getShippingSettings
+      responses:
+        '200':
+          description: ''
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    $ref: '#/components/schemas/shippingSettings'
+                  meta:
+                    type: object
+                    properties: {}
+      description: |-
+        Get shipping settings.
+      tags:
+        - Shipping Settings
+      summary: Get Shipping Settings
+    post:
+      operationId: updateShippingSettings
+      parameters:
+        - $ref: '#/components/parameters/ContentType'
+      responses:
+        '200':
+          description: ''
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    $ref: '#/components/schemas/shippingSettings'
+                  meta:
+                    type: object
+                    properties: {}
+      summary: Update Shipping Settings
+      description: |-
+        Updates shipping settings.
+      tags:
+        - Shipping Settings
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/shippingSettings'
 components:
   parameters:
     store_hash:
@@ -281,7 +334,7 @@ components:
           schema:
             $ref: '#/components/schemas/error_Full'
     405_Method_Not_Allowed:
-      description: The resource was found, but doesnʼt support the request method. Issued when either a specific method isnʼt yet implemented on a resource, or the resource doesnʼt support the method at all. For example, a PUT request to `/orders` is invalid, but a PUT request to `/orders/{order_id}` is valid. 
+      description: The resource was found, but doesnʼt support the request method. Issued when either a specific method isnʼt yet implemented on a resource, or the resource doesnʼt support the method at all. For example, a PUT request to `/orders` is invalid, but a PUT request to `/orders/{order_id}` is valid.
       content:
         application/json:
           schema:
@@ -323,7 +376,6 @@ components:
       title: customsInformationRequest
       description: Data about the customs information object.
       type: object
-      x-internal: false
       properties:
         product_id:
           description: The product ID to which the customs information data applies.
@@ -355,11 +407,23 @@ components:
         - commodity_description
         - international_shipping
         - hs_codes
+    shippingSettings:
+      type: object
+      properties:
+        checkout:
+          type: object
+          properties:
+            country_list_strategy:
+              description: The approach for displaying the list of countries at checkout.
+              type: string
+              enum:
+                - DISPLAY_ALL_COUNTRIES
+                - DISPLAY_ONLY_SHIPPABLE_COUNTRIES
+              example: DISPLAY_ALL_COUNTRIES
     customsInformation:
       title: customsInformation
       description: Data about the customs information object.
       type: object
-      x-internal: false
       properties:
         product_id:
           description: The ID of the product which the customs information data will apply to.
@@ -415,7 +479,6 @@ components:
         CA: '508313'
         US: '641000'
         AU: '817355'
-      x-internal: false
     metaCollection:
       title: metaCollection
       description: Meta data relating to pagination.
@@ -459,11 +522,9 @@ components:
                   type: string
                   description: Query string appended to the resource to show the current page.
                   example: '?limit=1&page=2'
-      x-internal: false
     error_Full:
       title: Error
       description: ''
       type: object
       properties: {}
       additionalProperties: true
-      x-internal: false


reference/shipping.v3.yml
```diff
@@ -397,6 +465,24 @@ components:
           readOnly: true
           format: date-time
           example: 2022-09-21T14:15:00+00:00
+    checkoutSettings:
+      title: checkoutSettings
+      description: Data about the checkout settings object.
```
benpratt77, line 470: This seems like a redundant comment.
Is there some value we can add here to be more descriptive of what one might find in these settings?

reference/shipping.v3.yml
```diff
@@ -142,6 +143,65 @@ paths:
...
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/checkoutSettings'
+                  meta:
+                    $ref: '#/components/schemas/metaCollection'
+      description: |-
+        Get checkout settings.
+
+        ```http
+        GET /shipping/settings
```
bc-andreadao, line 171: It'd be great you removed this example in the description because it'll be redundant when it's live on the site

shipping settings

reference/shipping.v3.yml
```diff
@@ -397,6 +465,24 @@ components:
           readOnly: true
           format: date-time
           example: 2022-09-21T14:15:00+00:00
+    checkoutSettings:
+      title: checkoutSettings
+      description: Data about the checkout settings object.
+      type: object
+      x-internal: false
```
bc-andreadao, line 472:
```suggestion
```

I believe the x-internal flags aren't actually needed anymore (we used to use a different platform that needed them)

reference/shipping.v3.yml
```diff
@@ -355,6 +415,14 @@ components:
         - commodity_description
         - international_shipping
         - hs_codes
+    checkoutSettings_request:
+      title: checkoutSettingsRequest
+      description: Data about the checkout settings object.
+      type: object
+      x-internal: false
```
bc-andreadao, line 422:
```suggestion
```
The x-internal flags aren't needed anymore

reference/shipping.v3.yml
```diff
@@ -397,6 +465,24 @@ components:
...
+      title: checkoutSettings
+      description: Data about the checkout settings object.
+      type: object
+      x-internal: false
+      properties:
+        checkout:
+          $ref: '#/components/schemas/countryListStrategy'
+    countryListStrategy:
+      title: countryListStrategy
+      description: Data about the strategy of listing countries at checkout
```
bc-andreadao, line 478: This might be redundant with the description for `country_list_strategy`

reference/shipping.v3.yml
```diff
@@ -397,6 +465,24 @@ components:
...
+      type: object
+      x-internal: false
+      properties:
+        country_list_strategy:
+          description: The strategy of how countries list is displayed at checkout.
```
bc-andreadao, line 483:
```suggestion
          description: The approach for displaying the list of countries at checkout.
```

reference/shipping.v3.yml
jvpcode: Perhaps adding info that it's shipping specific checkout settings?

reference/shipping.v3.yml
```diff
@@ -397,6 +465,24 @@ components:
...
+          type: string
```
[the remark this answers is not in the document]
payaalaee: Thanks Andrew, adde in the values🍻

reference/shipping.v3.yml
payaalaee: Thanks gents, Is `Checkout settings related to shipping.` descriptive enough?

reference/shipping.v3.yml
payaalaee: Thanks Andrea, I've removed this.🍻

reference/shipping.v3.yml
payaalaee: Thanks, I've removed all references to `x-internal` in this file.

reference/shipping.v3.yml
payaalaee: Thanks, updated🙏

reference/shipping.v3.yml
payaalaee: Thanks Andrea, I believe this is now more descriptive.

bc-andreadao: @payaalaee I'm actually thinking we could remove the sentence since the meaning is the same as the description for `country_list_strategy`. How does that sound?

reference/shipping.v3.yml
```diff
@@ -142,6 +143,61 @@ paths:
                     CA: '508313'
                     AU: '817355'
                     ALL: '501000'
+
+  '/shipping/settings':
+    parameters:
+      - $ref: '#/components/parameters/Accept'
+
+    get:
+      operationId: getCheckoutSettings
```
agchan12, line 152: I think I prefer the language of shippingSettings rather than checkoutSettings

reference/shipping.v3.yml
```diff
@@ -142,6 +143,61 @@ paths:
...
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/checkoutSettings'
+                  meta:
+                    $ref: '#/components/schemas/metaCollection'
+      description: |-
+        Get checkout settings.
+      tags:
+        - Checkout Settings
+      summary: Get Checkout Settings
```
agchan12, line 171: Remember that this is a shipping/settings endpoint. It currently just happens that the only relevant setting is applicable to checkout, but this is still shipping-settings.

reference/shipping.v3.yml
payaalaee: Thanks Andrea, it's removed!🙏

reference/shipping.v3.yml
payaalaee: Thanks Andrew, that's right, I forgot checkout is only one setting of shipping. I've updated all to `Shipping Settings`🍻

bc-andreadao (approved):

jvpcode (approved):

theromulans (approved):

agchan12 (approved):

reference/shipping.v3.yml
```diff
@@ -3,13 +3,14 @@ info:
...
-    url: 'https://www.bigcommerce.com' 
-    email: support@bigcommerce.com 
+  termsOfService: 'http://www.bigcommerce.com/terms'
+  contact:
+    name: BigCommerce
+    url: 'https://www.bigcommerce.com'
+    email: support@bigcommerce.com
 tags:
   - name: Customs Information
+  - name: Shipping Settings
```
theromulans, line 13: 🍻 This seems like a strange tag - is this the usual approach? 

I could imagine `Shipping` being a relevant tag, less sure about `Shipping Settings`. Did you have guidance here @bc-andreadao?

reference/shipping.v3.yml
```diff
@@ -355,11 +410,17 @@ components:
         - commodity_description
         - international_shipping
         - hs_codes
+    shippingSettings_request:
+      title: shippingSettingsRequest
+      description: Data about the shipping settings object.
+      type: object
+      properties:
+        checkout:
```
theromulans, line 418: Is this not `checkout_settings`?

reference/shipping.v3.yml
```diff
@@ -397,6 +458,24 @@ components:
           readOnly: true
           format: date-time
           example: 2022-09-21T14:15:00+00:00
+    shippingSettings:
+      title: shippingSettings
+      description: Shipping settings.
+      type: object
+      properties:
+        checkout:
```
theromulans, line 466: See above 🙏

reference/shipping.v3.yml
```diff
@@ -397,6 +458,24 @@ components:
...
+          $ref: '#/components/schemas/countryListStrategy'
```
theromulans, line 467: I'm also unsure about this - is this actually `checkoutSettings`, one property of which (for now) is `country_list_strategy`?

That is, if we add another property to this area (e.g., our out of zone delivery message) it would appear as another property within `checkoutSettings`?

reference/shipping.v3.yml
payaalaee: Thanks Brett, I originally had checkout settings there, but then @agchan12 suggested this is in the shipping domain with a single checkout setting in it for now. What are your thoughts @agchan12 ?

agchan12: I think @theromulans was more suggesting that the tag of `Shipping Settings` might be irrelevant and can be replaced to something more like `Shipping` instead.

For the tag recommendation though, they are used to group API endpoints logically, making the API documentation easier to navigate and understand. I am ok with keeping things generic and having the `shipping` tag separate. Keep in mind multiple tags can be assigned to an endpoint too.

theromulans: The list already includes `Customs Information` - this is why I'm asking for guidance from @bc-andreadao about what these values drive. Clearly our new Shipping Settings endpoint has no relevance to Customs Information either.

vborinsk: This doc somewhat explains how tags work - https://idratherbewriting.com/learnapidoc/pubapis_openapi_step7_tags_object.html
in summary - the global tags group lists all the defined tags on the current doc page, each tag underneath a `path` section denotes that path section tag (which drives the nav menu and headings I believe, once the doc site is built), if that makes sense.

theromulans: Indeed, if this is driving the navigation and headings on our docs site then there's no question about the most appropriate tag being `Shipping Settings` for our two new operations.

For example, the current `Customs Information` outcome:

Screenshot 2024-09-10 at 4 26 32 PM

reference/shipping.v3.yml
bc-andreadao: @payaalaee Could you put some screenshots of the intended request and response bodies in the PR description? The other option is to enable the feature on my store 

Currently, this is how the GET response body is documented (not sure if it's right)
GET response

This is how the POST request and response is documented: 
POST request and response

reference/shipping.v3.yml
bc-andreadao: Yes, the tags create the dividers on the site (as shown in @theromulans screenshot). 

[Line 170](https://github.com/bigcommerce/docs/blob/d08d3a38e0eb2eda3eb08a5fd049568f3f2d871a/reference/shipping.v3.yml#L170) and [Line193](https://github.com/bigcommerce/docs/blob/d08d3a38e0eb2eda3eb08a5fd049568f3f2d871a/reference/shipping.v3.yml#L193) make it so that the `GET /shipping/settings` endpoint and `POST /shipping/settings` endpoint appear under the "Shipping Settings" divider. 

The global tags ([line 11](https://github.com/bigcommerce/docs/blob/d08d3a38e0eb2eda3eb08a5fd049568f3f2d871a/reference/shipping.v3.yml#L11)) is just a cumulative list of all the tags that we put in the doc. If we don't include it, we'll get a linting error.

I don't have an opinion about whether or not it should be "Shipping Settings" or "Settings"

bc-andreadao: One minor detail is that the tag name also appears in the URL. 
URL

reference/shipping.v3.yml
theromulans: Thanks Andrea, Shipping Settings looks good then, thanks for confirming how the tags are used.

Curiously, do you know if a tag's `description` is used (if we define one)?

reference/shipping.v3.yml
theromulans: Thanks @bc-andreadao, Shipping Settings looks good then, thanks for confirming how the tags are used.

Curiously, do you know if a tag's description is used (if we define one)?

reference/shipping.v3.yml
```diff
@@ -142,6 +143,61 @@ paths:
...
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    type: array
```
theromulans, line 185: As also shown in @bc-andreadao's screenshot - this isn't an array of shipping settings, this is just an object. Providing the actual request and response payloads should highlight this inconsistency.

It looks like this is a copy paste from customs information, which is indeed an array of various customs information entities.

reference/shipping.v3.yml
theromulans: To clarify, I understand the outcome is as closed to the expected, my suggestion is renaming the `ref`. The scenario of adding another field to the `checkout` object above highlights the incorrect abstraction.

reference/shipping.v3.yml
```diff
@@ -397,6 +458,24 @@ components:
           readOnly: true
           format: date-time
           example: 2022-09-21T14:15:00+00:00
+    shippingSettings:
```
theromulans, line 461: Another item from our equivalent [Tax Settings implementation](https://github.com/bigcommerce/docs/blob/d08d3a38e0eb2eda3eb08a5fd049568f3f2d871a/reference/tax_settings.v3.yml#L54) is that we re-use the same "Tax_Settings" schema object for both request and response, given there is no difference and we expect no difference.

reference/shipping.v3.yml
bc-andreadao: I actually am not sure about tag descriptions (only some parts of the OAS3 file spec are supported on our site, so it may or may not work)

theromulans (changes requested):
@payaalaee Are you able to provide the API specification in the PR description so reviewers can more easily confirm the outcome matches the documentation?

reference/shipping.v3.yml
payaalaee: Thanks Andrea for the explanation and clarification on this. Staying with shipping settings here.

reference/shipping.v3.yml
payaalaee: Thanks Brett, yeah right my bad - I've updated the type to object. Gotta scan better next time when I copy paste from some other existing code😅

reference/shipping.v3.yml
payaalaee: Thanks, the entire code in this part was wrong and I've updated it similar to the other usages. There is now only one `shippingSettings` schema with the correct format being used across get and post request and response samples. I believe this is looking good now. Thanks for your comments. I'm also attaching a screenshot of the API request and response.

reference/shipping.v3.yml
payaalaee: Thanks Brett, I've removed `shippingSettingsRequest` and now only `shippingSettings` is used across this file. Looking all correct and clean😅🍻

reference/shipping.v3.yml
theromulans: Wonderful @payaalaee, thank you!

reference/shipping.v3.yml
theromulans: Really nice, cheers!

reference/shipping.v3.yml
```diff
@@ -355,11 +407,17 @@ components:
         - commodity_description
         - international_shipping
         - hs_codes
+    shippingSettings:
+      title: shippingSettingsRequest
+      description: Data about the shipping settings object.
+      type: object
+      properties:
+        checkout:
+          $ref: '#/components/schemas/countryListStrategy'
```
theromulans, line 416: Looks like one final thing to resolve here, which relates to avoiding mixing `countryListStrategy` with `checkout` for our `$ref` value. As I mentioned elsewhere, I think the ultimate solution is this...

```suggestion
    shippingSettings:
      type: object
      properties:
        checkout:
          type: object
          properties:
            country_list_strategy:
              description: The approach for displaying the list of countries at checkout.
              type: string
              enum:
                - DISPLAY_ALL_COUNTRIES
                - DISPLAY_ONLY_SHIPPABLE_COUNTRIES
              example: DISPLAY_ALL_COUNTRIES
```

Worth noting is that I've removed title and description from `shippingSettings` also, since they don't add any meaningful value (and it matches our [Tax_Settings](https://github.com/bigcommerce/docs/blob/d08d3a38e0eb2eda3eb08a5fd049568f3f2d871a/reference/tax_settings.v3.yml#L54) approach).

In addition to this change, we then remove the `countryListStrategy` definition (which is actually a definition for the _checkout_ object) on line 458 since it's now unused.

Another benefit here is that all the information about `shippingSettings` is in the one schema definition rather than jumbled up among the YML definitions.

payaalaee: Oh I see - Thanks mate, I've updated this as per your suggestion. I now have a better idea how to do docs tasks😅🙏🍻

theromulans: Thank you sir 🚀

theromulans (approved):
👏 Cheers!

bc-andreadao (approved):
Thanks @payaalaee !

PR #491, opened 2024-09-03
Update babel dependency in stencil react foundation doc

becomevocal:
## What changed?
* Updated the Babel plugin dependency that's recommended in the React foundation doc for Stencil, since it has moved into a new package since the doc was original created

## Release notes draft
* Stencil React foundation doc has been updated to work with the latest Babel plugin packages. Now the guidance works for React 18.


bc-traciporter (approved):
Approved

PR #503, opened 2024-09-06
Document proposed shipping methods v3 API

bc-luke:
Draft shipping methods API to support new shipping promotions. This doesn't need to be public but perhaps we will choose to release it publicly. To be discussed.

image

## What changed?

Document the parts of a shipping methods API that would be necessary to support the new shipping promotions UI including: 
* Filter by name, channel ID and zone
* Optionally include zone information (zone name and ID) and channel IDs
* Method name
* Method ID

## Release notes draft

* 

## Anything else?

ping {names}

### DIFF
--- a/reference/shipping.v3.yml
+++ b/reference/shipping.v3.yml
@@ -94,8 +94,7 @@ paths:
         - in: query
           name: include
           description: |-
-            * `shipping_zone` - shipping zone information
-            * `shipping_methods.channel_ids` - IDs of the associated channels
+            * `shipping_zone` - Include shipping zone information with each shipping method.
           style: form
           explode: false
           schema:
@@ -467,7 +466,7 @@ components:
               type: string
               example: "Thailand"
         channel_ids:
-          description: Channel information. This is available only when "include=shipping_zone" is present in the querystring.
+          description: List of channel IDs for which this method is applicable.
           type: array
           items:
             type: integer
--- a/reference/shipping.v3.yml
+++ b/reference/shipping.v3.yml
@@ -76,7 +76,7 @@ paths:
               type: integer
         - in: query
           name: 'name:like'
-          description: 'A name by which to filter shipping methods For more information, see [Filtering](/docs/start/about/common-query-params).'
+          description: 'A name by which to filter shipping methods. For more information, see [Filtering](/docs/start/about/common-query-params).'
           schema:
             type: string
         - in: query
@@ -87,7 +87,7 @@ paths:
             type: integer
         - in: query
           name: cursor
-          description: 'Encoded cursor'
+          description: 'A cursor pointing to the desired set of entities.'
           required: false
           schema:
             type: string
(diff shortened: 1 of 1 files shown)


reference/shipping.v3.yml
```diff
@@ -355,6 +440,41 @@ components:
...
+          description: The unique identifier of the shipping method.
+          type: integer
+          format: int32
+          example: 2
+        name:
+          description: The human-readable name of the shipping method.
+          type: string
+          example: "Courier delivery"
+        shipping_zone:
+          description: Zone information. This is available only when "include=shipping_zone" is present in the querystring.
```
theromulans, line 459: Seems like a sensible optional include 👌

reference/shipping.v3.yml
```diff
@@ -355,6 +440,41 @@ components:
         - commodity_description
         - international_shipping
         - hs_codes
+    shippingMethods:
+      title: shippingMethods
+      description: Data about the customs information object.
```
theromulans, line 445: Oops!

reference/shipping.v3.yml
```diff
@@ -355,6 +440,41 @@ components:
...
+      type: object
+      x-internal: false
```
theromulans, line 447: What does this represent?

reference/shipping.v3.yml
```diff
@@ -355,6 +440,41 @@ components:
...
+              description: The unique identifier of the shipping zone.
+              type: integer
+              format: int32
+              example: 2
+            name:
+              description: The human-readable name of the shipping zone.
+              type: string
+              example: "Thailand"
+        channel_ids:
+          description: Channel information. This is available only when "include=shipping_zone" is present in the querystring.
```
theromulans, line 472: Would we not want to always include the Channel IDs? Well, perhaps only when the store has multiple channels?

reference/shipping.v3.yml
vborinsk: Let's make it public, I have seen at least a few requests from agencies to have an API to query all methods without having to iterate over all the zones.

reference/shipping.v3.yml
```diff
@@ -20,6 +20,91 @@ servers:
...
+          name: cursor
+          description: 'Encoded cursor'
+          required: false
+          schema:
+            type: string
+        - in: query
+          name: include
+          description: |-
+            * `shipping_zone` - shipping zone information
+            * `shipping_methods.channel_ids` - IDs of the associated channels
```
agchan12, line 98:
```suggestion
          description: |-
            * `shipping_zone` - Include shipping zone information with each shipping method.
            * `shipping_methods.channel_ids` - Include the IDs of the associated channels for each shipping method.

```

reference/shipping.v3.yml
bc-luke: Yes probably a reasonable expecation. Updated!

reference/shipping.v3.yml
bc-luke: 🔧Fixed

reference/shipping.v3.yml
bc-luke: Fixed

reference/shipping.v3.yml
bc-luke: In that case, it might make sense to add more properties so we have feature parity with V2.

PR #551, opened 2024-09-27
Issue-522

bc-traciporter:
#Issue-522

## What changed?
Correct the first Update an Order request example

## Release notes draft

Big Fix

## Anything else?

ping {names}


donald-nguyen-bc (approved):
👍  thanks for the update Traci 🙇

PR #554, opened 2024-09-30
chore(shipping): SHIPPING-3046 Update V2 shipping docs with channels

payaalaee:
# [SHIPPING-3046](https://bigcommercecloud.atlassian.net/browse/SHIPPING-3046?atlOrigin=eyJpIjoiN2QxOWM2MDhkYWQ4NGNmN2JiZDhjYWVkZGE1MjI1MjgiLCJwIjoiaiJ9)

## What changed?

- Update V2 shipping document introducing channels

## Release notes draft
As a part of MSF project and introducing channels, we want to update our docs to reflect the new feature.
`channels` filed is the list of channels associated with a method.

## Anything else?
N/A

ping @bigcommerce/team-shipping @bc-andreadao 

[SHIPPING-3046]: https://bigcommercecloud.atlassian.net/browse/SHIPPING-3046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/shipping.v2.yml
+++ b/reference/shipping.v2.yml
@@ -733,6 +733,7 @@ paths:
                       handling_fees:
                         fixed_surcharge: "0"
                       is_fallback: false
+                      channels: [2, 3]
                     - id: 2
                       name: Pickup In Store
                       type: total
@@ -741,6 +742,7 @@ paths:
                       handling_fees:
                         fixed_surcharge: "0"
                       is_fallback: false
+                      channels: [1, 3]
                     - id: 3
                       name: Ship by Weight
                       type: weight
@@ -764,6 +766,7 @@ paths:
                       handling_fees:
                         fixed_surcharge: "0"
                       is_fallback: false
+                      channels: [1]
                     - id: 4
                       name: UPS®
                       type: upsready
@@ -785,6 +788,7 @@ paths:
                       handling_fees:
                         fixed_surcharge: "0"
                       is_fallback: false
+                      channels: [1]
                 Example 2:
                   value:
                     - id: 23225205
@@ -795,6 +799,7 @@ paths:
                       handling_fees:
                         fixed_surcharge: "0"
                       is_fallback: false
+                      channels: [1, 2]
                     - id: 35889344
                       name: Lorem Excepteur esse
                       type: canadapost
@@ -803,6 +808,7 @@ paths:
                       handling_fees:
                         percentage_surcharge: "0"
                       is_fallback: false
+                      channels: [1]
       operationId: getShippingZoneMethods
     post:
       description: |-
@@ -834,7 +840,7 @@ paths:
         #### USPS by Endicia Object Properties
         
         | Property | Type | Values |
-        |:---------|:-----|:-------|
+        | - | - | - |
         | delivery_services | array | PriorityExpress, PriorityMailExpressInternational, FirstClassPackageInternationalService, Priority, PriorityMailInternational, First, ParcelSelect, MediaMail |
         | packaging_type | array | FlatRateLegalEnvelope, FlatRatePaddedEnvelope, Parcel, SmallFlatRateBox, MediumFlatRateBox, LargeFlatRateBox, FlatRateEnvelope, RegionalRateBoxA, RegionalRateBoxB |
         |show_transit_time | boolean | true, false |
@@ -869,7 +875,7 @@ paths:
         #### FedEx Object Properties
         
         | Property | Type | Values |
-        |:---------|:-----|:-------|
+        | - | - | - |
         | delivery_services | array | PRIORITY_OVERNIGHT, STANDARD_OVERNIGHT, FIRST_OVERNIGHT, FEDEX_2_DAY, FEDEX_EXPRESS_SAVER, INTERNATIONAL_PRIORITY, INTERNATIONAL_ECONOMY, INTERNATIONAL_FIRST, FEDEX_1_DAY_FREIGHT, FEDEX_2_DAY_FREIGHT, FEDEX_3_DAY_FREIGHT, FEDEX_GROUND, GROUND_HOME_DELIVERY, INTERNATIONAL_PRIORITY_FREIGHT, INTERNATIONAL_ECONOMY_FREIGHT, EUROPE_FIRST_INTERNATIONAL_PRIORITY |
         | dropoff_type | string | REGULAR_PICKUP, REQUEST_COURIER, DROP_BOX, BUSINESS_SERVICE_CENTER, STATION |
         | packaging_type | string | FEDEX_ENVELOPE,  FEDEX_PAK, FEDEX_BOX, FEDEX_TUBE, FEDEX_10KG_BOX, FEDEX_25KG_BOX, YOUR_PACKAGING |
@@ -907,7 +913,7 @@ paths:
         #### UPS Ready Object Properties
         
         | Property | Type | Values |
-        |:---------|:-----|:-------|
+        | - | - | - |
         | delivery_services | array | 2nd_Day_Air, 2nd_Day_Air_AM, 3_Day_Select, Expedited, Express, Express_Plus, Express_Saver, Express_Early_AM, Ground, Next_Day_Air, Next_Day_Air_Early_AM, Next_Day_Air_Saver, Saver, Standard, Today_Dedicated_Courier, Today_Express, Today_Express_Saver, Today_Intercity, Today_Standard, Worldwide_Expedited, Worldwide_Express, Worldwide_Express_Plus, Worldwide_Express_Saver, Worldwide_Saver |
         | destination_type | string | residential, business |
         | packing_method | string | separate, combined |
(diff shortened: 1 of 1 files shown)


reference/shipping.v2.yml
```diff
@@ -733,6 +733,7 @@ paths:
                       handling_fees:
                         fixed_surcharge: "0"
                       is_fallback: false
+                      channels: [2,3]
```
benpratt77, line 736: [2, 3]
here and below
or is the lack of spacing a standard here?

reference/shipping.v2.yml
```diff
@@ -1452,6 +1469,12 @@ paths:
                     description: Whether or not this shipping zone is the fallback if all others are not valid for the order.
                     example: false
                     type: boolean
+                  channels:
+                    description: List of channels associated to a method. On new method creation all available channels are associated by default.
```
benpratt77, line 1473: When creating a new method, all available channels are associated by default.

reference/shipping.v2.yml
payaalaee: Hmm thanks Ben, I was mistaken it would be different with the docs, and though in other areas in the docs there are ones with no spac, I think we can apply styling best practice here as pioneers😆🍻

bc-andreadao (approved):

bc-andreadao:
Thanks @payaalaee . I created [PR 556](https://github.com/bigcommerce/docs/pull/556) so we can talk about the use cases & limitations in the doc, if you could review/approve

reference/shipping.v2.yml
```diff
@@ -1377,7 +1393,8 @@ paths:
                         "shipping_cost": 0
                     }
                 ]
-            }
+            },
+            "channels": [1, 3]
```
theromulans, line 1397: 🍻 I think it'd actually be better to avoid updating most of our existing examples.

At the moment channel management isn't an overly common use case. In this specific example, the core example isn't speaking to channels.

> This example sets free shipping above a certain order total

Hence I think it's a distraction to include the channels field here, and probably in all of the existing examples.

If we reconsider this from the ground up, focusing on the specific use cases where we are creating a simple shipping method for a channel, then it might be better. That is, as part of the description specifying why the `channels` field has been used and what it's effect is.

reference/shipping.v2.yml
```diff
@@ -1452,6 +1469,12 @@ paths:
                     description: Whether or not this shipping zone is the fallback if all others are not valid for the order.
                     example: false
                     type: boolean
+                  channels:
```
theromulans, line 1472: Q: is this field only present when the merchant has more than one storefront channel?

reference/shipping.v2.yml
```diff
@@ -2337,6 +2369,12 @@ components:
           description: Whether or not this shipping method is a fallback method used when advanced shipping rules are unavailable.
           example: false
           type: boolean
+        channels:
+          description: List of channels associated to a method. When creating a new method, all available channels are associated by default.
+          example: [1, 3]
+          type: array
+          items:
+            type: integer
```
theromulans, line 2377: Would it be beneficial to indicate that this is an optional field?

reference/shipping.v2.yml
payaalaee: Thanks Brett, you're absolutely right. With the post and put examples I've introduced this new one dedicated to channels.

This particular line though falls under the get, and channels field will be present in every call, so I'm keeping the channels in this area.

This looks clearer now.🍻🙏

reference/shipping.v2.yml
payaalaee: Like I mentioned above, this field is always present regardless of number of store channels. It's only absent in case of experiment off which would be irrelevant in this case, since this doc will be updated only once MSF is live.🍻

reference/shipping.v2.yml
payaalaee: Thanks mate, I've added `(optional)` to the description for both instances.🍻

reference/shipping.v2.yml
```diff
@@ -1070,7 +1076,7 @@ paths:
         |:-----|:-----|:------------|
         | rate | number | Flat rate per item. |
         
-        Ezample request body:
+        Example request body:
```
theromulans, line 1079: Heh, nice.

reference/shipping.v2.yml
```diff
@@ -1189,6 +1195,25 @@ paths:
...
         }
         ```
+        
+        ### Channels
+
+        Example request body: 
+
+        ```json
+        {
+          "name": "Method associated to channels 1,3",
```
theromulans, line 1205:
```suggestion
          "name": "Method associated to channels 1, 3",
```

🍹

reference/shipping.v2.yml
```diff
@@ -1189,6 +1195,25 @@ paths:
...
+          "settings": {
+            "rate": 5
+          },
+          channels: [1, 3]
+        }
+        ```
+
+        | Property | Type | Description |
+        |:---------|:-----|:------------|
+        | channels | array | channels associated with the method as an array of integers |
```
theromulans, line 1216:
```suggestion
        | channels | array | Channels associated with the method as an array of integers |
```

reference/shipping.v2.yml
```diff
@@ -1400,6 +1434,25 @@ paths:
             "shipping_cost": 8
         }
         ```
+        
+        ### Channels
+
+                Example request body:
```
theromulans, line 1440: Looks like this block has an extra level of indentation compared to the `Range Object – Properties` example above?

reference/shipping.v2.yml
```diff
@@ -1400,6 +1434,25 @@ paths:
...
         }
         ```
+        
+        ### Channels
+
+                Example request body: 
+
+                ```json
+                {
+                  "name": "Method associated to channels 1,3",
```
theromulans, line 1444:
```suggestion
                  "name": "Method associated to channels 1, 3",
```

🍹

reference/shipping.v2.yml
```diff
@@ -1400,6 +1434,25 @@ paths:
...
+                  "settings": {
+                    "rate": 5
+                  },
+                  channels: [1, 3]
+                }
+                ```
+
+                | Property | Type | Description |
+                |:---------|:-----|:------------|
+                | channels | array | channels associated with the method as an array of integers |
```
theromulans, line 1455:
```suggestion
                | channels | array | Channels associated with the method as an array of integers |
```

reference/shipping.v2.yml
```diff
@@ -1400,6 +1434,25 @@ paths:
...
+                  "type": "peritem",
+                  "settings": {
+                    "rate": 5
+                  },
+                  channels: [1, 3]
+                }
+                ```
+
+                | Property | Type | Description |
+                |:---------|:-----|:------------|
```
theromulans, line 1454:
```suggestion
                | - | - | - |
```

I think we can replace the above with the simplified definition, a little easier to maintain, since this updated definition doesn't have any relation to the header row's column name length. As seen in `Range Object – Properties` example.

reference/shipping.v2.yml
payaalaee: 😅

reference/shipping.v2.yml
payaalaee: Thanks mate, updated all 3 instances of this. 🍻

reference/shipping.v2.yml
payaalaee: Same as this one, updated the others as well. 🍻

reference/shipping.v2.yml
payaalaee: Thanks for the suggestion Brett, yeah exactly. I've updated all the instances since we're here. 🍻

theromulans: Lovely sir, thanks!

theromulans (approved):

benpratt77 (approved):

reference/shipping.v2.yml
```diff
@@ -1639,7 +1700,24 @@ paths:
...
+          "settings": {
+          "rate": 5
+          },
+          channels: [1, 3]
+        }
+        ```
+
+        | Property | Type | Description |
+        | - | - | - |
+        | channels | array | Channels associated with the method as an array of integers |
```
bc-andreadao, line 1720:
```suggestion
        | channels | array | Channels associated with the method as an array of integers. |
```

reference/shipping.v2.yml
```diff
@@ -1189,6 +1195,25 @@ paths:
...
+          "settings": {
+            "rate": 5
+          },
+          channels: [1, 3]
+        }
+        ```
+
+        | Property | Type | Description |
+        | - | - | - |
+        | channels | array | Channels associated with the method as an array of integers |
```
bc-andreadao, line 1216:
```suggestion
        | channels | array | Channels associated with the method as an array of integers. |
```

bc-andreadao (approved):
Looks good, left minor punctuation suggestions to keep the style consistent

bc-andreadao (approved):

PR #564, opened 2024-10-07
DEVDOCS-6113: [update] matches aftership list

bc-traciporter:
# [DEVDOCS-6113]

## What changed?
Updated our tracking carrier list with list from aftership 

## Release notes draft

* 

## Anything else?

ping {names}

[DEVDOCS-6113]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-6113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bc-traciporter:
@Ben Can you also review this PR?

benpratt77:
You should check this against https://github.com/bigcommerce/shipping-service/blob/865c454d588a0e38740d800e07c6cbd6bb8fd6fe/src/Services/TrackingLink/ExternalCarriers.php#L4 That is our current list of providers that are supported from within the code base.
This is what we use under the hood to figure out whether we reach out to aftership.

If there are missing names in that file it is worth raising a ticket for us to update.

benpratt77:
> @ben Can you also review this PR?

might be the wrong ben 🤣

bc-traciporter:
> > @ben Can you also review this PR?
> 
> might be the wrong ben 🤣

Sorry about that

bc-traciporter:
> You should check this against https://github.com/bigcommerce/shipping-service/blob/865c454d588a0e38740d800e07c6cbd6bb8fd6fe/src/Services/TrackingLink/ExternalCarriers.php#L4 That is our current list of providers that are supported from within the code base. This is what we use under the hood to figure out whether we reach out to aftership.
> 
> If there are missing names in that file it is worth raising a ticket for us to update.

[benpratt77](https://github.com/benpratt77)

I don't have access to that link. I get a 404.

theromulans:
One thing just observed is that `directfreight-au-ref`, for example, is not actually a tracking carrier string we support - suggesting perhaps there's a more modern variant we need here?

Outside of that, I do regard continuing to update these docs manually as a bandaid solution. Would there be an option for something that is generated by engineering, directly referencing our actual source of truth that is active in production?

PR #626, opened 2024-11-03
Update b2b bearer token to follow OAS spec

becomevocal:
## What changed?
Updated BearerToken security scheme for B2B storefront specs to follow OAS spec. This is needed so they render under the Authorization section in the right way.


bc-traciporter (approved):
approved

PR #640, opened 2024-11-08
DEVDOCS-6122: [update] change limit from 10 to 3

bc-traciporter:
# [DEVDOCS-6122]

## What changed?
Added concurrency rate limits to 3.

## Release notes draft

Enhancement: Added concurrency rate limits to Update Customers and Upsert Customer attribute values endpoint
.

* 

## Anything else?

ping {names}

[DEVDOCS-6122]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-6122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/customers.v3.yml
+++ b/reference/customers.v3.yml
@@ -334,7 +334,7 @@ paths:
         * origin_channel_id
 
         **Limits**
-        Limit of 10 customers per call.
+        Limit of 3 customers per call.
         
         **Notes**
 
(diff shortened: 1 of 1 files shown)


bc-RyanGross:
Thanks for working on this Traci! The changes needed were relevant to total concurrent API requests to these endpoints rather than the amount of Customer records that could be upserted in a single request.

So for both of these endpoints:
https://developer.bigcommerce.com/docs/rest-management/customers#update-customers
https://developer.bigcommerce.com/docs/rest-management/customers/attribute-values#upsert-customer-attribute-values

We need to specify that there is a concurrent API request rate limit of 3 concurrent requests (this would be new content that does not currently exist in the documentation for both of the endpoints specified above). The amount of Customer records per Upsert request does not need to change, it is still 10 per request as currently documented.

bc-RyanGross (approved):
Looks great!

PR #641, opened 2024-11-08
DEVDOCS-5967: [update] update discount_rules array options

bc-traciporter:
# [DEVDOCS-5967]

## What changed?
Add discount array options for category-level, product-level and Storewide-level discounts.

## Release notes draft

Bug Fix

* 

## Anything else?

ping {names}

[DEVDOCS-5967]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/customers.v2.yml
+++ b/reference/customers.v2.yml
@@ -1220,32 +1220,107 @@ components:
         category_access:
           $ref: '#/components/schemas/categoryAccessLevel_Full'
         discount_rules:
-          type: array
-          description: A collection of discount rules that are automatically applied to customers who are members of the group.
-          items:
-            type: object
-            properties:
-              type:
-                type: string
-                enum:
-                  - price_list
-                  - all
-                  - category
-                  - product
-              method:
-                type: string
-                enum:
-                  - percent
-                  - fixed
-                  - price
-              amount:
-                type: string
-                description: A float that specifies the value applied to the price modified. (Float, Float as String, Integer)
-                example: "5.0000"
-              price_list_id:
-                type: integer
-                description: If a customer group is assigned to a price list,`method` and `amount` are not shown. `type` and `price_list_id` are returned.
-                example: 3
+          oneOf:
+          - type: array
+            description: A collection of discount rules that are automatically applied to customers who are members of the group.
+            items:
+              type: object
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - price_list
+                    - all
+                    - category
+                    - product
+                method:
+                  type: string
+                  enum:
+                    - percent
+                    - fixed
+                    - price
+                amount:
+                  type: string
+                  description: A float that specifies the value applied to the price modified. (Float, Float as String, Integer)
+                  example: "5.0000"
+                price_list_id:
+                  type: integer
+                  description: If a customer group is assigned to a price list,`method` and `amount` are not shown. `type` and `price_list_id` are returned.
+                  example: 3
+          - type: array
+            description: A collection of category-level discount rules that are automatically applied to customers who are members of the group.
+            items:
+              type: object
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - price_list
+                    - all
+                    - category
+                    - product
+                method:
+                  type: string
+                  enum:
+                    - percent
+                    - fixed
+                    - price
+                amount:
+                  type: string
+                  description: A float that specifies the value applied to the price modified. (Float, Float as String, Integer)
+                  example: "5.0000"
+                category_id:
+                  type: integer
+                  description: If a customer group is assigned to a category, `method` and `amount` are not shown. `type` and `category_id` are returned.
+                  example: 3
+          - type: array
+            description: A collection of product-level discount rules that are automatically applied to customers who are members of the group.
+            items:
+              type: object
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - price_list
+                    - all
+                    - category
+                    - product
+                method:
+                  type: string
+                  enum:
+                    - percent
+                    - fixed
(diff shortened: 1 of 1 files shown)


reference/customers.v2.yml
```diff
@@ -1220,32 +1220,107 @@ components:
...
+            items:
+              type: object
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - price_list
+                    - all
+                    - category
+                    - product
```
isaacchen-bc, line 1261: since we have `category_id` in this array, then I think here it would have to be `category`?

reference/customers.v2.yml
```diff
@@ -1220,32 +1220,107 @@ components:
...
```
isaacchen-bc, line 1287: Similar with the above comment, since we have product_id in this array, then I think the type here would have to be product?

reference/customers.v2.yml
```diff
@@ -1220,32 +1220,107 @@ components:
...
```
isaacchen-bc, line 1235: Should we remove `product` and `category` here because these two cases are covered by the below? I guess if one does not provide `product_id` or `category_id`, then even putting product or category as type here, it wouldn't work 🤔

reference/customers.v2.yml
```diff
@@ -1220,32 +1220,103 @@ components:
...
+          - type: array
+            description: A collection of product-level discount rules that are automatically applied to customers who are members of the group.
+            items:
+              type: object
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - price_list
+                    - all
```
isaacchen-bc, line 1282: shouldn't we remove these two because there is `product_id` under this array type?

reference/customers.v2.yml
```diff
@@ -1220,32 +1220,103 @@ components:
...
+          - type: array
+            description: A collection of category-level discount rules that are automatically applied to customers who are members of the group.
+            items:
+              type: object
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - price_list
+                    - all
```
isaacchen-bc, line 1257: similar with https://github.com/bigcommerce/docs/pull/641/files#r1837343493, should we remove price_list and all in this array type?

reference/customers.v2.yml
```diff
@@ -1220,32 +1220,103 @@ components:
...
+                    - category
+                    - product
```
isaacchen-bc, line 1309: I think we can only leave `all` here? because others (`price_list`, `category` and `product`) are already covered by the above cases

reference/customers.v2.yml
```diff
@@ -1220,32 +1220,103 @@ components:
...
+          - type: array
+            description: A collection of discount rules that are automatically applied to customers who are members of the group.
+            items:
+              type: object
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - price_list
+                    - all
```
isaacchen-bc, line 1233: should we remove `all` here? because we have `price_list_id` here? will need to confirm with pricing team on this one.

reference/customers.v2.yml
isaacchen-bc, line 1: My comments on customerGroup_Full are all applicable to customerGroup_Post_Put 🙇

bc-traciporter:
HI @isaacchen-bc I made the updates you mentioned above. Can you confirm the changes are correct?

isaacchen-bc (approved):
Thanks @bc-traciporter 🙇 Looks good to me! 
I guess let's also confirm with pricing team for this as well.

reference/customers.v2.yml
```diff
@@ -1220,32 +1220,95 @@ components:
...
+          - type: array
+            description: A collection of discount rules that are automatically applied to customers who are members of the group.
+            items:
+              type: object
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - price_list
+                method:
```
megdesko, line 1233: There's no method or amount here in the API, just the price list ID.

reference/customers.v2.yml
```diff
@@ -1349,32 +1412,95 @@ components:
...
+                amount:
+                  type: string
+                  description: A float that specifies the value applied to the price modified. (Float, Float as String, Integer)
+                  example: "5.0000"
+                product_id:
+                  type: integer
+                  description: If a customer group is assigned to a product,`method` and `amount` are not shown. `type` and `product` are returned.
+                  example: 3
+          - type: array
+            description: A collection of storewide discount rules that are automatically applied to customers who are members of the group.
```
megdesko, line 1486: There can only be a single storewide discount rule on a customer group

bc-traciporter:
Hi @megdesko thank you for the feedback. Can you review my changes?

megdesko:
@bc-traciporter This still seems wrong to me...  it's (pseudocode)
```
oneof
array
- contains a single price list (so type = price_list, has the price_list_id parameter)

array
- contains the complete set of all discounts (now this is tricky, because the payload for the discounts differs by type... ) 
```
If the client sends updates with different arrays containing sitewide/ product/ category they're going to wipe the other rules they already have.  

It's documented in this PR as `oneof` 4 different arrays - price lists, sitewide discounts, category discounts, product discounts... and this is misleading.

I'm not sure there's a good way to communicate this via this type of document, TBH.

bc-traciporter:
@megdesko Instead of using OneOf, I can use anyOf. Will that work?

bc-traciporter:
Too many conflicts, will close PR and create a new one.

PR #662, opened 2024-11-15
LOCAL-2603 Fix geo swagger definition

funivan:
## What changed?

1. Set default value for accept and content-type. Make swagger file consistent
2. When using autogenerated TS library we have an issue with types inconsistency. `country_id` is integer. Int is convertable to string, so there should be no issues

## Release notes draft
None

## Anything else?
None


bc-traciporter (approved):
Approved

PR #668, opened 2024-11-19
DEVDOCS-6151: [update] removed required label where necessary

bc-traciporter:
# [DEVDOCS-6151]

## What changed?
The Update Products (Batch) and Create a Product endpoints share the product_Base schema. Therefore, I created a separate request schema for each endpoint. 

## Release notes draft

bug fix: Removed the incorrect required flag for the Update Products (batch) to prevent confusion.

* 

## Anything else?

ping {names}

[DEVDOCS-6151]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-6151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/catalog/products_catalog.v3.yml
```diff
@@ -7289,11 +7281,6 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/productVariant_Full'
-      required:
```
bc-romanliukshyn, line 7292: @bc-traciporter does this update mean that the `required` label will be gone for the Create Product sections?

bc-traciporter: Hi @bc-romanliukshyn Yes, I added the requirement in the description field since the schema is shared by both the Create Products and Update Products endpoint.  Having the required tag caused confusion with the Update Product endpoint since it is not needed.

bc-traciporter: Let me know if you have a better solution? We could always create a separate schema if you think that is a better solution.

bc-romanliukshyn: In my opinion, the `required` label serves an important purpose and enhances readability more effectively than a comment beside the field. Therefore, I believe we should keep it, even if it requires additional effort to restructure the schema.

bc-traciporter: OK, no problem. I will create separate schemas for POST and PUT requests.

bc-romanliukshyn (dismissed):

PR #642, opened 2024-11-11
doc(checkout): CHECKOUT-8739 update doc to reflect new feature of order level fees.

bc-shawnwang:
# [CHECKOUT-8739](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-8739)

## What changed?
v3 checkout api:
- ~~Create Fee~~ (see "Updates after change to bulk APIs")
- ~~Update Fee~~ (see "Updates after change to bulk APIs")
- ~~Delete Fee~~ (see "Updates after change to bulk APIs")
- response of checkout object added applied fees (fees:)

storefront checkout api:
- response of checkout object added applied fees (fees:)

-----

<details open>
<summary> <h3> Updates after change to bulk APIs, 2024-11-28 </h3> </summary>

on commit: https://github.com/bigcommerce/docs/pull/642/commits/5dd480c9c0107b6b6df0c5e735630fb2376222b9

v3 checkout api changes:

Have 3 endpoints to support bulk operations on fees:
- POST v3/checkouts/:checkoutId/fees
- PUT v3/checkouts/:checkoutId/fees
- DELETE v3/checkouts/:checkoutId/fees

Removed 2 endpoints:
- PUT v3/checkouts/:checkoutId/fees/:feeId
- DELETE v3/checkouts/:checkoutId/fees/:feeId

Response of v3 checkout and storefront checkout have nothing changed.

### Overall 
image

### bulk POST
image

### bulk PUT
Screenshot 2024-11-28 at 10 11 24 AM

### bulk DELETE
image

</details>

## Release notes draft

* The newly-released custom order level fees on the checkout is now available to use. Now, you’ll be able to apply and manage the checkout's order level fee via v3 management api, and fetch any checkout's applied fees on storefront checkout api and v3 management checkout api. 

## Anything else?

ping @bigcommerce/team-checkout @bigcommerce/dev-docs 

[CHECKOUT-8739]: https://bigcommercecloud.atlassian.net/browse/CHECKOUT-8739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #681, opened 2024-11-25
feat(orders): ORDERS-6751 Doc update with order level fee in order apis

dishadutta26:
# ORDERS-6751

## What changed?
Orders v2 api
* Get Fee
* Create Fee
* Update Fee

## Release notes draft
* The newly-released custom order level fees on the orders is now available to use. Now, you’ll be able to apply and manage the order level fee via orders v2 api, and fetch any applied fees on an order.

## Anything else?
ping @bigcommerce/team-orders

### DIFF
--- a/reference/orders.v2.oas2.yml
+++ b/reference/orders.v2.oas2.yml
@@ -101,7 +101,7 @@ paths:
         After the update, the PUT request clears all discounts and promotions applied to the changed order line items. Since the order data syncs with other ERP systems, like Amazon or eBay, the updated order returns to the default setting, removing any applied discounts.
         
         To update order fees, include the fee id in the request body along with all relevant fee fields. Fees not included will be deleted. Fees with an id will be updated, and fees without an id will be created as new.
-        Note: Sub-resources like products in the /v2/orders PUT request behave like PATCH, updating only the provided fields. Fees, however, follow standard PUT semantics—fees in the request body will fully replace existing ones. To retain an existing fee, include it in the body with its associated id.
+        Note: Sub-resources like products in the /v2/orders PUT request behave like PATCH, updating only the provided fields. Fees, however, follow standard PUT semantics fees in the request body will fully replace existing ones. To retain an existing fee, include it in the body with its associated id.
         
         To learn more about creating or updating orders, see [Orders Overview](/docs/store-operations/orders).
       summary: Update an Order
@@ -160,11 +160,17 @@ paths:
                   fees:
                     - id: 1
                       type: custom_fee
-                      display_name_customer: ABC Retail Delivery Fee
-                      source: v2
+                      display_name_customer: A Fee
+                      source: AA
                       cost_ex_tax: 12.3000
                       cost_inc_tax: 12.3000
                       tax_class_id: 22
+                    - id: 2
+                      type: custom_fee
+                      display_name_customer: B Fee
+                      source: AA
+                      cost_ex_tax: 8.0000
+                      cost_inc_tax: 12.0000
               Adding an existing product to order:
                 value:
                   products:
@@ -194,70 +200,55 @@ paths:
                           value: 12
                       price_inc_tax: 12.45
                       price_ex_tax: 10.12
-              Adding a fee to the order (fee provided without an ID):
+              Adding a new fee (no ID) while retaining existing fees (with ID):
                 value:
                   fees:
                     - type: custom_fee
                       display_name_customer: C Fee
-                      source: v2
+                      source: AA
                       cost_ex_tax: '22.5000'
                       cost_inc_tax: '25.0000'
-                      tax_class_id: 22
                     - id: 1
                       type: custom_fee
                       display_name_customer: A Fee
-                      source: v2
+                      source: AA
                       cost_exc_tax: '12.3000'
-                      cost_inc_tax: '12.30000'
+                      cost_inc_tax: '12.3000'
                       tax_class_id: 22
                     - id: 2
                       type: custom_fee
                       display_name_merchant: B Fee
-                      source: v2
+                      source: AA
                       cost_ex_tax: '4.0000'
                       cost_tax: '8.0000'
-                      tax_class_id: 0
+                      tax_class_id: null
               Updating order fees (all fees provided with IDs and attributes):
                 value:
                   fees:
                     - id: 3
                       type: custom_fee
-                      display_name_customer: Retail Delivery Fee
-                      source: v2
-                      cost_ex_tax: '22.5000'
+                      display_name_customer: C Fee Updated
+                      source: AA
+                      cost_ex_tax: '20.0000'
                       cost_inc_tax: '25.0000'
-                      tax_class_id: 22
                     - id: 1
                       type: custom_fee
-                      display_name_customer: A Fee
-                      source: v2
-                      cost_exc_tax: '12.3000'
-                      cost_inc_tax: '14.30000'
-                      tax_class_id: 22
+                      display_name_customer: A Fee Updated
+                      source: AA
+                      cost_exc_tax: '10.0000'
+                      cost_inc_tax: '15.0000'
+                      tax_class_id: 1
                     - id: 2
                       type: custom_fee
-                      display_name_merchant: B Fee
-                      source: v2
+                      display_name_merchant: B Fee Updated
+                      source: AA
                       cost_ex_tax: '5.0000'
                       cost_tax: '8.0000'
-                      tax_class_id: 0
+                      tax_class_id: 2
               Deleting a fee from the order (fee missing from the payload will be removed):
                 value:
                   fees:
-                    - id: 1
-                      type: custom_fee
-                      display_name_customer: A Fee
-                      source: v2
-                      cost_exc_tax: '12.3000'
-                      cost_inc_tax: '14.30000'
-                      tax_class_id: 22
-                    - id: 2
-                      type: custom_fee
-                      display_name_merchant: B Fee
-                      source: v2
-                      cost_ex_tax: '5.0000'
-                      cost_tax: '8.0000'
-                      tax_class_id: 0
+                    -
 
         required: true
       responses:
@@ -398,14 +389,6 @@ paths:
                     - product_id: 118
                       quantity: 1
                       variant_id: 93
-                  fees:
-                    - type: custom_fee
-                      display_name_customer: ABC Retail Delivery Fee
-                      display_name_merchant: ABC Retail Delivery Fee
-                      source: v2
-                      cost_ex_tax: 12.3000
-                      cost_inc_tax: 12.3000
-                      tax_class_id: 22
               Custom Product:
                 value:
                   status_id: 0
@@ -531,27 +514,26 @@ paths:
                           value: '180'
                         - id: 230
                           value: '192'
-              Adding a fee to the order (fee provided without an ID):
+              Create an order with fees (include the fees object as shown with rest of the payload):
                 value:
                   fees:
                     - type: custom_fee
                       display_name_customer: D Fee
-                      source: v2
+                      source: AA
                       cost_ex_tax: '22.5000'
                       cost_inc_tax: '25.0000'
                       tax_class_id: 22
                     - type: custom_fee
                       display_name_customer: E Fee
-                      source: v2
+                      source: AA
                       cost_exc_tax: '12.3000'
-                      cost_inc_tax: '12.30000'
-                      tax_class_id: 22
+                      cost_inc_tax: '12.3000'
+                      tax_class_id: null
                     - type: custom_fee
                       display_name_merchant: F Fee
-                      source: v2
+                      source: AA
                       cost_ex_tax: '4.0000'
                       cost_tax: '8.0000'
-                      tax_class_id: 0
         required: true
       responses:
         '200':
@@ -1186,22 +1168,22 @@ paths:
                         type: custom_fee
                         display_name_customer: A Fee
                         display_name_merchant: A Fee
-                        source: v2
+                        source: AA
                         base_cost: '12.3000'
                         cost_exc_tax: '12.3000'
-                        cost_inc_tax: '12.30000'
+                        cost_inc_tax: '12.3000'
                         cost_tax: '0.0000'
                         tax_class_id: 22
                       - id: 2
                         type: custom_fee
                         display_name_customer: B Fee
                         display_name_merchant: B Fee
-                        source: v2
+                        source: AA
                         base_cost: '4.0000'
                         cost_ex_tax: '4.0000'
                         cost_inc_tax: '12.0000'
                         cost_tax: '8.0000'
-                        tax_class_id: 0
+                        tax_class_id: null
         '404':
           $ref: '#/components/responses/404_Resp'
       operationId: getOrderFees
@@ -1880,22 +1862,22 @@ components:
                     type: custom_fee
                     display_name_customer: A Fee
                     display_name_merchant: A Fee
-                    source: v2
+                    source: AA
                     base_cost: '12.3000'
                     cost_exc_tax: '12.3000'
-                    cost_inc_tax: '12.30000'
+                    cost_inc_tax: '12.3000'
                     cost_tax: '0.0000'
                     tax_class_id: 22
                   - id: 2
                     type: custom_fee
                     display_name_customer: B Fee
                     display_name_merchant: B Fee
-                    source: v2
+                    source: AA
                     base_cost: '4.0000'
                     cost_ex_tax: '4.0000'
                     cost_inc_tax: '12.0000'
                     cost_tax: '8.0000'
-                    tax_class_id: 0
+                    tax_class_id: null
             'Multiple Items':
               value:
                 id: 247
@@ -2095,22 +2077,22 @@ components:
                     type: custom_fee
                     display_name_customer: A Fee
                     display_name_merchant: A Fee
-                    source: v2
+                    source: AA
                     base_cost: '12.3000'
                     cost_exc_tax: '12.3000'
-                    cost_inc_tax: '12.30000'
+                    cost_inc_tax: '12.3000'
                     cost_tax: '0.0000'
                     tax_class_id: 22
                   - id: 2
                     type: custom_fee
                     display_name_customer: B Fee
                     display_name_merchant: B Fee
-                    source: v2
+                    source: AA
                     base_cost: '4.0000'
                     cost_ex_tax: '4.0000'
                     cost_inc_tax: '12.0000'
                     cost_tax: '8.0000'
-                    tax_class_id: 0
+                    tax_class_id: null
             'Multiple Items':
               value:
                 id: 247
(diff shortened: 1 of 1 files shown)


reference/orders.v2.oas2.yml
```diff
@@ -99,6 +100,8 @@ paths:
 
         After the update, the PUT request clears all discounts and promotions applied to the changed order line items. Since the order data syncs with other ERP systems, like Amazon or eBay, the updated order returns to the default setting, removing any applied discounts.
         
+        To update the fees for an order, include the fee `id` in the request body. The body should include all relevant fields, with either the updated values or the existing values you wish to retain. Fees not included in the payload will be deleted. Fees provided in the request will be updated, and fees without an ID will be created as new.
```
donald-nguyen-bc, line 103: 🍹 Probably we can include the consequence for fields not included in the request body.
```suggestion
        To update the fees for an order, include the fee `id` in the request body. The body should include all relevant fee fields, with either the updated values or the existing values you wish to retain (missing fee fields will be reset to empty value). Fees not included in the payload will be deleted. Fees provided with an ID in the request will be updated, and fees without an ID will be created as new.
```

reference/orders.v2.oas2.yml
```diff
@@ -1058,6 +1079,363 @@ paths:
...
+                        cost_inc_tax: '12.0000'
+                        cost_tax: '8.0000'
+                        tax_class_id: 0
+        '404':
+          $ref: '#/components/responses/404_Resp'
+      operationId: getOrderFees
+      description: 'Get all fees for an order. '
+      parameters:
+        - $ref: '#/components/parameters/order_id_path'
+  '/orders/{order_id}?include=fees':
```
donald-nguyen-bc, line 1125: this probably is not a new endpoint, we should update the existing one and update `include` param instead

reference/orders.v2.oas2.yml
```diff
@@ -1058,6 +1079,363 @@ paths:
...
+      summary: Update fees for an order
+      tags:
+        - Order Fees
+      parameters:
+        - $ref: '#/components/parameters/ContentType'
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/orderFees'
```
donald-nguyen-bc, line 1192: probably not all the fields in this object are supported in PUT/POST request

reference/orders.v2.oas2.yml
```diff
@@ -1058,6 +1079,363 @@ paths:
...
+                        cost_inc_tax: '12.0000'
+                        cost_tax: '8.0000'
+                        tax_class_id: 0
+        '404':
+          $ref: '#/components/responses/404_Resp'
+      operationId: getOrderFees
+      description: 'Get all fees for an order. '
+      parameters:
+        - $ref: '#/components/parameters/order_id_path'
+    put:
```
donald-nguyen-bc, line 1180: do we need to update order_Put, order_Post schemas as well for the 'PUT /orders/{order_id}' and 'POST /orders' endpoints ?

reference/orders.v2.oas2.yml
dishadutta26: For missing fields, fee will actually throw 400
image

reference/orders.v2.oas2.yml
dishadutta26: Thanks updated, I had added in the existing ones and created new ones as well to explain it explicitly for fees but lint check did not allow query params anyways

reference/orders.v2.oas2.yml
WillemHoman, line 103: It's probably worth mentioning something along the lines of 


Most sub-resources like `products`  included in the `/v2/orderds` PUT request body act more like PATCH in that they retain the existing values and only update those included in the request body. 

Fees has more standard PUT semantics in that the fees in the request body will completely replace any existing fees on the order. 
To retain an existing fee, include it in the PUT request body with it's associated `id` etc.

WillemHoman:
It's worth mentioning that for `tax_class_id` if it is absent or `null` it will be set to the default `tax_class_id` of `0`.

reference/orders.v2.oas2.yml
```diff
@@ -5415,3 +6140,51 @@ components:
...
+          description: The fee cost value including tax. (Float, Float-As-String, Integer)
+          example: 10.0000
+          type: number
+        cost_tax:
+          description: The tax amount on the fee cost. (Float, Float-As-String, Integer) 
+                       NOTE - At least two of the following fields (cost_ex_tax, cost_inc_tax, and cost_tax) must be included in the request.
+          example: 1.5000
+          type: number
+        tax_class_id:
+          description: A unique numeric identifier for the tax class object. The value will be 0 if automatic tax is enabled or if the default tax class is applied.
```
WillemHoman, line 6188: It's worth mentioning that for `tax_class_id` if it is absent or `null` it will be set to the default `tax_class_id` of `0`.

WillemHoman (approved):
👏

reference/orders.v2.oas2.yml
dishadutta26: Updated schemas, thanks

reference/orders.v2.oas2.yml
dishadutta26: Added separate schema for all

reference/orders.v2.oas2.yml
```diff
@@ -99,6 +100,9 @@ paths:
 
         After the update, the PUT request clears all discounts and promotions applied to the changed order line items. Since the order data syncs with other ERP systems, like Amazon or eBay, the updated order returns to the default setting, removing any applied discounts.
         
+        To update order fees, include the fee id in the request body along with all relevant fee fields. Fees not included will be deleted. Fees with an id will be updated, and fees without an id will be created as new.
+        Note: Sub-resources like products in the /v2/orders PUT request behave like PATCH, updating only the provided fields. Fees, however, follow standard PUT semantics—fees in the request body will fully replace existing ones. To retain an existing fee, include it in the body with its associated id.
```
WillemHoman, line 104: 🍹 `semantics-fees` needs spaces around the `-`

WillemHoman:
Updates look good @dishadutta26

reference/orders.v2.oas2.yml
```diff
@@ -182,10 +194,75 @@ paths:
                           value: 12
                       price_inc_tax: 12.45
                       price_ex_tax: 10.12
+              Adding a fee to the order (fee provided without an ID):
```
donald-nguyen-bc, line 197:
```suggestion
              Adding a new fee to the order (fee provided without an ID) and retain existing fees (fees with an ID):
```

reference/orders.v2.oas2.yml
```diff
@@ -182,10 +194,75 @@ paths:
...
+                      source: v2
+                      cost_ex_tax: '22.5000'
+                      cost_inc_tax: '25.0000'
+                      tax_class_id: 22
+                    - id: 1
+                      type: custom_fee
+                      display_name_customer: A Fee
+                      source: v2
+                      cost_exc_tax: '12.3000'
+                      cost_inc_tax: '12.30000'
```
donald-nguyen-bc, line 211:
```suggestion
                      cost_inc_tax: '12.3000'
```

reference/orders.v2.oas2.yml
```diff
@@ -444,10 +531,31 @@ paths:
                           value: '180'
                         - id: 230
                           value: '192'
+              Adding a fee to the order (fee provided without an ID):
```
donald-nguyen-bc, line 534:
```suggestion
              Create order with fees:
```

reference/orders.v2.oas2.yml
```diff
@@ -444,10 +531,31 @@ paths:
...
+                      source: v2
+                      cost_exc_tax: '12.3000'
+                      cost_inc_tax: '12.30000'
+                      tax_class_id: 22
+                    - type: custom_fee
+                      display_name_merchant: F Fee
+                      source: v2
+                      cost_ex_tax: '4.0000'
+                      cost_tax: '8.0000'
+                      tax_class_id: 0
```
donald-nguyen-bc, line 554: seems like this payload is not enough for POST /orders request, probably we can remove this example ?

reference/orders.v2.oas2.yml
```diff
@@ -1058,6 +1166,48 @@ paths:
...
+                  value:
+                    fees:
+                      - id: 1
+                        type: custom_fee
+                        display_name_customer: A Fee
+                        display_name_merchant: A Fee
+                        source: v2
+                        base_cost: '12.3000'
+                        cost_exc_tax: '12.3000'
+                        cost_inc_tax: '12.30000'
```
donald-nguyen-bc, line 1192:
```suggestion
                        cost_inc_tax: '12.3000'
```

reference/orders.v2.oas2.yml
```diff
@@ -1514,121 +1667,640 @@ components:
...
+                  resource: /orders/247/coupons
+                external_id: null
+                external_merchant_id: null
+                tax_provider_id: BasicTaxProvider
+                store_default_currency_code: USD
+                store_default_to_transactional_exchange_rate: '1'
+                custom_status: Awaiting Fulfillment
+                customer_locale: en
+                external_order_id: external-order-id
+    orderPut_Resp:
```
donald-nguyen-bc, line 1992: just curious why do we need this new `orderPut_Resp` along with `order_Resp` ?

reference/orders.v2.oas2.yml
```diff
@@ -1514,121 +1667,640 @@ components:
...
+                  resource: /orders/247/coupons
+                external_id: null
+                external_merchant_id: null
+                tax_provider_id: BasicTaxProvider
+                store_default_currency_code: USD
+                store_default_to_transactional_exchange_rate: '1'
+                custom_status: Awaiting Fulfillment
+                customer_locale: en
+                external_order_id: external-order-id
+            Fee Added:
```
donald-nguyen-bc, line 2207: Do we need these examples: "Fee added", "Fee updated", and "Fee deleted" ? They look quite the same, and they don't include the other fields on the "Order response" ? I'm not sure if it's helpful for client to have all 3 those examples.

reference/orders.v2.oas2.yml
```diff
@@ -1514,121 +1667,640 @@ components:
...
+                    type: custom_fee
+                    display_name_customer: B Fee
+                    display_name_merchant: B Fee
+                    source: v2
+                    base_cost: '13.0000'
+                    cost_ex_tax: '5.0000'
+                    cost_inc_tax: '13.0000'
+                    cost_tax: '8.0000'
+                    tax_class_id: 0
+    orderPost_Resp:
```
donald-nguyen-bc, line 2296: same as orderPut_Resp, do we need this to be separated ?

reference/orders.v2.oas2.yml
```diff
@@ -4196,6 +4922,11 @@ components:
           description: The status ID of the order.
         billing_address:
           $ref: '#/components/schemas/billingAddress_Resp'
+        fees:
+          type: array
+          items:
+            allOf:
+              - $ref: '#/components/schemas/orderFees_Resp'
```
donald-nguyen-bc, line 4929: probably we can remove the allOf

reference/orders.v2.oas2.yml
```diff
@@ -5415,3 +6157,153 @@ components:
...
+          description: The display name of the fee targeting customers.
+          example: Package Protection Insurance
+          type: string
+        display_name_merchant:
+          description: The display name of the fee targeting shoppers.
+            NOTE - At least one of the following fields (display_name_customer, display_name_merchant) must be included in the request.
+          example: Package Protection Fee
+          type: string
+        source:
+          description: TThe source of the request.
```
donald-nguyen-bc, line 6183:
```suggestion
          description: The source of the request.
```

reference/orders.v2.oas2.yml
```diff
@@ -5415,3 +6157,153 @@ components:
...
+            - type: number
+        cost_inc_tax:
+          description: The fee cost value including tax. (Float, Float-As-String, Integer)
+          example: 10.0000
+          oneOf:
+            - type: string
+            - type: number
+        cost_tax:
+          description: The tax amount on the fee cost. (Float, Float-As-String, Integer)
+            NOTE - At least two of the following fields (cost_ex_tax, cost_inc_tax, and cost_tax) must be included in the request.
```
donald-nguyen-bc, line 6207: probably we can remove this line for response schema ?

reference/orders.v2.oas2.yml
```diff
@@ -5415,3 +6157,153 @@ components:
...
+          example: 1.5000
+          oneOf:
+            - type: string
+            - type: number
+        tax_class_id:
+          description: A unique numeric identifier for the tax class. It defaults to 0 if automatic tax is enabled, the default tax class is applied, or if tax_class_id is absent or null.
```
donald-nguyen-bc, line 6213: this seems to be not very clear, if `tax_class_id` is provided in payload and automatic tax is enabled, then what value will be assigned to that field ?

reference/orders.v2.oas2.yml
```diff
@@ -5415,3 +6157,153 @@ components:
...
+          description: The display name of the fee targeting customers.
+          example: Package Protection Insurance
+          type: string
+        display_name_merchant:
+          description: The display name of the fee targeting shoppers.
+            NOTE - At least one of the following fields (display_name_customer, display_name_merchant) must be included in the request.
+          example: Package Protection Fee
+          type: string
+        source:
+          description: TThe source of the request.
```
donald-nguyen-bc, line 6235:
```suggestion
          description: The source of the request.
```

donald-nguyen-bc, line 6235: Do we allow client to provide this value ? The sample data is "v2" which suggests that this value is auto assigned by our system when fee is created/update by v2 API, is it correct ?

reference/orders.v2.oas2.yml
```diff
@@ -5415,3 +6157,153 @@ components:
...
+            - type: number
+        cost_inc_tax:
+          description: The fee cost value including tax. (Float, Float-As-String, Integer)
+          example: 10.0000
+          oneOf:
+            - type: string
+            - type: number
+        cost_tax:
+          description: The tax amount on the fee cost. (Float, Float-As-String, Integer)
+            NOTE - At least two of the following fields (cost_ex_tax, cost_inc_tax, and cost_tax) must be included in the request.
```
donald-nguyen-bc, line 6252: do we need to mention the case where all those fields are provided and `cost_ex_tax + cost_tax != cost_inc_tax`

reference/orders.v2.oas2.yml
```diff
@@ -153,6 +157,14 @@ paths:
                         - id: 230
                           value: '192'
                   customer_locale: en
+                  fees:
+                    - id: 1
+                      type: custom_fee
+                      display_name_customer: ABC Retail Delivery Fee
+                      source: v2
+                      cost_ex_tax: 12.3000
```
bc-vincent-zhao, line 165: as we discussed offline, we should add some explanation to clearly call out that the `(cost_ex_tax, cost_inc_tax)` should have the correct values that reflects the tax rate referenced by the `tax_class_id` 

If the `tax_class_id` refers to a 10% tax rate, then the `(cost_ex_tax, cost_inc_tax)` should contain 10% tax when we do the subtraction. 

If the `tax_class_id` is omitted, the store configured "default tax class for fee" would be used and the cost values should match that.

If the user fails to provide accurate data here, downstream operation such as refund may have unintended surprises .

reference/orders.v2.oas2.yml
dishadutta26: The source, yes we allow to pass in, checkout allows too so could be any custom named source I guess

reference/orders.v2.oas2.yml
dishadutta26: will add something as Vincent suggested to avoid issues . As also mentioned in https://github.com/bigcommerce/bigcommerce/pull/60962

reference/orders.v2.oas2.yml
dishadutta26: Will have an updated description here based on our recent findings

reference/orders.v2.oas2.yml
donald-nguyen-bc: thanks for the info Disha, should we provide more realistic sample data for this field, atm it's "v2" and may cause confusion.

dishadutta26: Ok, using example similar to Checkout now https://github.com/bigcommerce/docs/pull/642/files#diff-2f53894a7f3c1f46d3d98dfd5e4bbb1e2f7a557aa8329ee806a96d047291f6b9R9227

reference/orders.v2.oas2.yml
```diff
@@ -99,6 +100,14 @@ paths:
...
         
+        To update order fees, include the fee id in the request body along with all relevant fee fields. Fees not included will be deleted. Fees with an id will be updated, and fees without an id will be created as new.
+        
+        **Notes**
+        
+        * Sub-resources like products in the /v2/orders PUT request behave like PATCH, updating only the provided fields. Fees, however, follow standard PUT semantics and fees in the request body will fully replace existing ones. 
+        To retain an existing fee, include it in the body with its associated id. 
+        * The values for cost_ex_tax, cost_inc_tax and cost_tax in the fees payload should reflect the tax rate associated with the tax_class_id. For a 10% tax rate, the difference between cost_inc_tax and cost_ex_tax should be 10%. If no tax_class_id is provided, the store's default "tax class for fee" will apply. Incorrect data may lead to issues in downstream operations like refunds.
```
donald-nguyen-bc, line 109: > Incorrect data may lead to issues in downstream operations like refunds

should we do the validation for this logic ? Otherwise it may lead to uncertainty when using the API because there's no guarantee that the data provided by client is correct and does not lead to problem in downstream operations ?

bc-vincent-zhao: it's actually hard to do real validation because that would involve calling real tax providers to calculate the tax to se if it matches with the provided value (and we currently only support basic tax but not automatic tax providers in v2 api)

That's the reason we want to call it out very explicitly so it's users' own responsibility to ensure the data is correct.

reference/orders.v2.oas2.yml
```diff
@@ -182,6 +205,56 @@ paths:
...
+                      type: custom_fee
+                      display_name_merchant: B Fee Updated
+                      source: AA
+                      cost_ex_tax: '5.0000'
+                      cost_tax: '8.0000'
+                      tax_class_id: 2
+              Deleting a fee from the order (fee missing from the payload will be removed):
+                value:
+                  fees:
+                    -
```
donald-nguyen-bc, line 256: if the fees is empty then this operation probably is "delete all"

donald-nguyen-bc (approved):

donald-nguyen-bc (dismissed):
👏

reference/orders.v2.oas2.yml
```diff
@@ -1278,6 +1426,8 @@ components:
         * `consignments` - include the response returned from the request to the `/orders/{order_id}/consignments` endpoint.
         
         * `consignments.line_items` - include the response returned from the request to the `/orders/{order_id}/products` endpoint in consignments. This implies `include=consignments`.
+
+        * `fees` - include the response returned from the request to the `/orders/{order_id}/fees` endpoint. This implies `include=fees`.
```
bc-luke, line 1430:
```suggestion
        * `fees` - include the response returned from the request to the `/orders/{order_id}/fees` endpoint.
```

I believe it's explicit in this case vs above where including line items will also include consignments.

donald-nguyen-bc (approved):

PR #721, opened 2024-12-20
DEVDOCS-6175: [update] update order of operations

bc-traciporter:
# [DEVDOCS-6175]

## What changed?
Moved customer group discounts and bulk pricing after Product modifier in the table.

## Release notes draft

Bug fix: Moved customer group discounts because they are shown before modifier rules when they apply after, as does bulk pricing.  

* 

## Anything else?

ping {names}

[DEVDOCS-6175]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-6175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/store-operations/pricing/calculations.mdx
+++ b/docs/store-operations/pricing/calculations.mdx
@@ -14,8 +14,8 @@ The table below lists each price type available on a product. The table is read 
 | Price list variant product pick list | The product pick list overrides price list bulk pricing and allows you to change the price when a pick list item is selected. | 
 | Product option set (**Deprecated**)| A set of product option facets that may alter the variant price. (This feature is deprecated.) |
 | Product modifier | Available in fixed ($5) or percentage (%10) amount that is added or removed from the total product price and overrides the price list variant product pick list. A modifier includes choices such as adding $5 for insurance. |
-| Customer group discount | Available in fixed ($5), relative (-$2), or percentage (-25%) amount that overrides the variant sale price. Might apply to one product, category, subcategory, or the entire store's products. |
-| Product bulk pricing | Available in fixed ($5), relative (-$2), or percentage (-25%) amount that overrides the customer group discount. Dependent on the total quantity of products, including SKUs added to the cart. |
+| Customer group discount | Available in fixed ($5), relative (-$2), or percentage (-25%) amount that overrides the variant sale price. Might apply to one product, category, subcategory, or the entire store's products. Customer group discounts do not apply to price list prices. |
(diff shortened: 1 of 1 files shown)


megdesko (commented):
@bc-traciporter  what do you think of splitting into two versions: one for price lists and one for non price lists requests?  Since you can't use both features at the same time, this may make more sense to users.

bc-traciporter:
@megdesko are you thinking of something like this?
Screenshot 2024-12-23 at 11 31 30 AM
Screenshot 2024-12-23 at 11 35 43 AM

docs/store-operations/pricing/calculations.mdx
```diff
@@ -8,14 +8,14 @@ The table below lists each price type available on a product. The table is read
...
 | Price list variant product pick list | The product pick list overrides price list bulk pricing and allows you to change the price when a pick list item is selected. | 
 | Product option set (**Deprecated**)| A set of product option facets that may alter the variant price. (This feature is deprecated.) |
 | Product modifier | Available in fixed ($5) or percentage (%10) amount that is added or removed from the total product price and overrides the price list variant product pick list. A modifier includes choices such as adding $5 for insurance. |
+| Customer group discount | Available in fixed ($5), relative (-$2), or percentage (-25%) amount that overrides the variant sale price. Might apply to one product, category, subcategory, or the entire store's products. |
+| Product bulk pricing | Available in fixed ($5), relative (-$2), or percentage (-25%) amount that overrides the customer group discount. Dependent on the total quantity of products, including SKUs added to the cart. |
```
megdesko, line 18:
```suggestion
| Customer group discount | Available in fixed ($5), relative (-$2), or percentage (-25%) amount that overrides the variant sale price. Might apply to one product, category, subcategory, or the entire store's products. Customer group discounts do not apply to price list prices. |
| Product bulk pricing | Available in fixed ($5), relative (-$2), or percentage (-25%) amount that overrides the customer group discount. Dependent on the total quantity of a single product added to the cart. Product bulk pricing will not apply for SKUs that have price list prices. |
```
@bc-traciporter  what do you think of something like this?  Want to make clear that these aren't in play when there's a price record available.

bc-traciporter: @megdesko that's fine. So we can keep the one table and add this text. I think it works.

megdesko (approved):

PR #723, opened 2024-12-22
Add target field object (with country_codes, subdivision_codes) to Tax Provider Connection API docs

theromulans:
# [TAX-2166]

## What changed?
* Add `target` object to Tax Provider Connection API docs. Includes `country_codes` and `subdivision_codes` fields.

## Release notes draft
- Allows users to inspect where their tax provider connection is active. Useful for determining if there are outstanding steps in the merchant's tax provider configuration.

[TAX-2166]: https://bigcommercecloud.atlassian.net/browse/TAX-2166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/tax.v3.yml
+++ b/reference/tax.v3.yml
@@ -207,4 +207,20 @@ components:
                 Describes whether the stored credentials are considered complete and configured, ready to be used for Tax Provider API requests.
 
                 Merchants may enable any **configured** tax provider for storefront tax quotation.
+            target:
+              type: object
+              description: The countries and subdivisions in which this tax provider connection is active.
+              properties:
+                country_codes:
+                  type: array
+                  items:
+                    type: string
+                  description: The list of country codes where the tax provider connection is active. ISO 3166-1 alpha-2.
+                  example: ['AU', 'US']
+                subdivision_codes:
+                  type: array
+                  items:
+                    type: string
+                  description: The list of subdivision codes where the tax provider connection is active. ISO 3166-2.
+                  example: ['AU-NSW', 'US-OH']
       x-internal: false


reference/tax.v3.yml
```diff
@@ -207,4 +207,20 @@ components:
...
+                  type: array
+                  items:
+                    type: string
+                  description: The list of country codes where the tax provider connection is active. ISO 3166-1 alpha-2 country code.
+                  example: ['AU', 'US']
+                subdivision_codes:
+                  type: array
+                  items:
+                    type: string
+                  description: The list of subdivision codes where the tax provider connection is active. ISO 3166-2 subdivision code, up to three alphanumeric characters, each with a ISO 3166-1 alpha-2 prefix.
```
agchan12, line 224: The description was a bit confusing around the `ISO-3166-2 subdivision code, up to three alphanumeric characters...`

What do you think about starting with describing the country prefix then describing the subdivision bit after?

theromulans: Good idea, it's out of order as it is now. Thanks!

theromulans: Actually, turns out this added description isn't needed - it's all part of the ISO 3166-2 standard.

That is `AU-NSW` is a complete ISO 3166-2 code, per the standard. I'll simplify the language.

reference/tax.v3.yml
```diff
@@ -207,4 +207,20 @@ components:
                 Describes whether the stored credentials are considered complete and configured, ready to be used for Tax Provider API requests.
 
                 Merchants may enable any **configured** tax provider for storefront tax quotation.
+            target:
+              type: object
+              description: A checkout session's active tax provider connection will be determined by the target data.
```
jvpcode, line 212: 🍹  This doesn't read well to me, the mention of a checkout session kinda threw me off. IMO something a little more direct might be easier to read? Trying to take the perspective of any of the tax integrators we work with.

I like the descriptions for `country_codes` and `subdivision_codes`, so something like
```suggestion
              description: The countries and subdivisions in which this tax provider is active
```

theromulans: Yeah, good call - I think something more straight forward will work better here.

One thing is that automatic tax providers are inherently only active in the checkout context, this doesn't really need to be described here.

declankirk (approved):
Makes sense to me 👍

bc-traciporter (approved):
Approved

PR #730, opened 2025-01-01
Add note about document submission operations

theromulans:
# [TAX-100]

## What changed?
* Added a note about document submission operations directly within our API reference.

We keep getting tax providers miss this key detail, so we're hoping this change helps reduce the number of these questions.

"Note" style is lifted from the nearby Tax Provider Connection API documentation.

Related docs: https://developer.bigcommerce.com/docs/rest-contracts/tax

[TAX-100]: https://bigcommercecloud.atlassian.net/browse/TAX-100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/integrations/tax.mdx
+++ b/docs/integrations/tax.mdx
@@ -33,14 +33,15 @@ Once your tax provider configuration is ready, we'll let you know through email.
 | Platform availability | Required | All stores, Limited availability | Whether your tax provider will be available to all stores or a limited number of stores. | `Limited availability` |
 | Supported / unsupported countries | Required | ISO 3166-1 alpha-2 | Comma separated ISO 3166-1 alpha-2 country codes for supported countries. You can also select to support to all countries, or all countries except a list of unsupported countries. | `US,CA,GB,FR,AU,NZ` |
 | **URLs** | | | | |
-| <sup>+</sup>Estimate URL | Required | URL | URL BigCommerce should use for Tax Provider API estimate requests. | `https://sampletax.example.com/tax/estimate` |
-| <sup>+</sup>Commit URL | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/commit` |
-| <sup>+</sup>Adjust URL | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/adjust` |
-| <sup>+</sup>Void URL | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/void` 
+| Estimate URL <sup>1</sup> | Required | URL | URL BigCommerce should use for Tax Provider API estimate requests. | `https://sampletax.example.com/tax/estimate` |
+| Commit URL <sup>1,2</sup> | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/commit` |
+| Adjust URL <sup>1,2</sup> | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/adjust` |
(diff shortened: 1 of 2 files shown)


jvpcode (commented):
This suggestion might be overkill, but what do you think about instead (or in addition) adding a similar note/s to the table with the details an integrator needs to provide us? i.e. the table here: https://developer.bigcommerce.com/docs/integrations/tax#sharing-provider-details-with-bigcommerce

My perspective is:
- Tax integrator wants to add a tax provider
- Looks at the details they need to provide
- Looks at the description for commit/void/adjust URLs
- Sees in the description for `adjust` (as an example)
```
URL BigCommerce should use for Tax Provider API quote requests. [Document submission](/docs/integrations/tax#document-submission) operations are only triggered when the *Submit Order Data* provider setting is enabled.
```

That way (hopefully) the integrator is hopefully already having that setting in mind when they send us the details to begin with

docs/integrations/tax.mdx
```diff
@@ -33,14 +33,15 @@ Once your tax provider configuration is ready, we'll let you know through email.
...
+| Adjust URL <sup>1</sup><sup>2</sup> | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/adjust` |
+| Void URL <sup>1</sup><sup>2</sup> | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/void`
 | **Testing** | | | | |
 | Partner sandbox store domain | Required | Domain name | Share your partner sandbox store for testing purposes prior to launching your tax provider. Learn how to [create a partner sandbox store](/docs/start/about/sandboxes). | `https://sampletax-test-store.mybigcommerce.com/` |
 
-<sup>+</sup> You have the option of supporting a URL that merchants can customize. See [Tax Profile](#tax-profile-optional).
+<sup>1</sup> You have the option of supporting a URL that merchants can customize. See [Tax Profile](#tax-profile-optional).
+<sup>2</sup> URL is only called when the **Submit Order Data** provider setting is enabled. See [Document Submission](#document-submission).
```
theromulans, line 44: @jvpcode Great suggestion sir, thanks. I've taken the approach of adding a superscript point to highlight this - WDYT? 🙏 Cheers.

docs/integrations/tax.mdx
```diff
@@ -33,14 +33,15 @@ Once your tax provider configuration is ready, we'll let you know through email.
 | Platform availability | Required | All stores, Limited availability | Whether your tax provider will be available to all stores or a limited number of stores. | `Limited availability` |
 | Supported / unsupported countries | Required | ISO 3166-1 alpha-2 | Comma separated ISO 3166-1 alpha-2 country codes for supported countries. You can also select to support to all countries, or all countries except a list of unsupported countries. | `US,CA,GB,FR,AU,NZ` |
 | **URLs** | | | | |
-| <sup>+</sup>Estimate URL | Required | URL | URL BigCommerce should use for Tax Provider API estimate requests. | `https://sampletax.example.com/tax/estimate` |
-| <sup>+</sup>Commit URL | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/commit` |
-| <sup>+</sup>Adjust URL | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/adjust` |
-| <sup>+</sup>Void URL | Optional | URL | URL BigCommerce should use for Tax Provider API quote requests. | `https://sampletax.example.com/doc/void`
```
theromulans, line 39: 👁️ I've long found it odd to have the superscript before the item being referenced, so I've also taken the opportunity to move it after.

docs/integrations/tax.mdx
jvpcode: applause

jvpcode (approved):

PR #744, opened 2025-01-13
Remove deprecated field, add new control panel price display strategy field

theromulans:
# [TAX-2184]

## What changed?
* Removed deprecated `show_inclusive_in_control_panel` field from Tax Settings API.
* Added `control_panel_price_display_strategy` field to Tax Settings API.

## Release notes draft
* Updated available Tax Settings API fields, new option available for controlling tax price display in the control panel.

[TAX-2184]: https://bigcommercecloud.atlassian.net/browse/TAX-2184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/tax_settings.v3.yml
+++ b/reference/tax_settings.v3.yml
@@ -124,9 +124,13 @@ components:
           type: object
           description: Settings that describe how prices display at the global level.
           properties:
-            show_inclusive_in_control_panel:
-              type: boolean
-              description: Whether to show prices as tax inclusive or tax exclusive in the BigCommerce control panel.
+            control_panel_price_display_strategy:
+              type: string
+              description: Whether to show prices as tax inclusive or tax exclusive in the BigCommerce control panel, or use the order's tax zone for price display.
+              enum:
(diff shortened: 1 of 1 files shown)


bc-traciporter (approved):
Approved

bc-andreadao:
@theromulans Would it make sense to have a `deprecated` flag next to the `show_inclusive_in_control_panel` field (instead of removing it)? 

The deprecated field would still show up in the example on the side (see screenshot), but I could change that by directly adding an full example to the yml file 

Screenshot 2025-01-13 at 1 59 40 PM

theromulans:
Good question @bc-andreadao. My thinking was that it actually causes an error using this field, so no point documenting it.

bc-andreadao:
Sounds good, thanks @theromulans

bc-andreadao (approved):

PR #761, opened 2025-01-17
Remove min/max property from the URL object

funivan:
# [DEVDOCS-]

## What changed?
Move min/max directive to the right property

## Release notes draft
None


bc-traciporter (dismissed):
Approved

reference/catalog/category-trees_catalog.v3.yml
```diff
@@ -906,8 +906,6 @@ components:
       description: |
         If not provided in POST request, the URL is autogenerated from the category name.
       type: object
-      minLength: 1
```
vil11, line 909: these limits must about the "path" property, the same as we have for deprecated Category API entity

vil11 (changes requested):

bc-romanliukshyn (approved):

reference/catalog/category-trees_catalog.v3.yml
funivan: fixed.

vil11: thnx a lot

vil11 (approved):

PR #769, opened 2025-01-22
chore(customers): CUST-3623 update customers api docs based on cursor pagination

isaacchen-bc:
# [CUST-3623](https://bigcommercecloud.atlassian.net/browse/CUST-3623)

## What changed?

* update v3 get customers api docs based on cursor pagination
* update v3 get ffvs api docs based on cursor pagination
* correct the value of meta for v3 post/put customers

#### get customers:
Screenshot 2025-01-23 at 11 15 39 AM
Screenshot 2025-01-23 at 11 15 48 AM
Screenshot 2025-01-23 at 11 16 39 AM
Screenshot 2025-01-23 at 11 16 50 AM

#### get ffvs:
Screenshot 2025-01-23 at 11 17 38 AM
Screenshot 2025-01-23 at 11 17 45 AM
Screenshot 2025-01-23 at 11 17 57 AM
Screenshot 2025-01-23 at 11 18 07 AM

## Release notes draft
* Feature: Cursor pagination is now available to Customers V3 - Get All Customers
* Feature: Cursor pagination is now available to Customers V3 - Get Customer Form Field Values
* Fixed: Corrected the value of meta for the response of Customers V3 - Create/Update Customers

[CUST-3623]: https://bigcommercecloud.atlassian.net/browse/CUST-3623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/customers.v3.yml
+++ b/reference/customers.v3.yml
@@ -52,7 +52,7 @@ paths:
       parameters:
         - name: page
           in: query
-          description: Page number (`page` will be ignored if `before` or `after` is provided in the request). For example `page=1`.
+          description: Page number (`page` will be ignored if you provide `before` or `after` in the request). For example `page=1`.
           schema:
             type: integer
         - in: query
@@ -1462,7 +1462,7 @@ paths:
             default: application/json
         - name: page
           in: query
-          description: Page number (`page` will be ignored if `before` or `after` is provided in the request). For example `page=1`.
+          description: Page number (`page` will be ignored if you provide `before` or `after` in the request). For example `page=1`.
           schema:
             type: integer
         - in: query
--- a/reference/customers.v3.yml
+++ b/reference/customers.v3.yml
@@ -210,12 +210,12 @@ paths:
               - 'date_modified:desc'
         - name: after
           in: query
-          description: 'The cursor reference of the last entry for the previous page. Use the `end_cursor` value from the previous response to get the next page (`end_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `after=eyJpZCI6MjA0fQ`. '
+          description: 'The cursor reference of the last entry for the previous page. Use the `end_cursor` value from the last response to get the next page (`end_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `after=eyJpZCI6MjA0fQ`. '
           schema:
             type: string
         - name: before
           in: query
-          description: 'The cursor reference of the first entry for the next page. Use the `start_cursor` value from the previous response to get the previous page (`start_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `before=eyJpZCI6MjA1fQ`.'
+          description: 'The cursor reference of the first entry for the next page. Use the `start_cursor` value from the last response to get the previous page (`start_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `before=eyJpZCI6MjA1fQ`.'
           schema:
             type: string
       responses:
@@ -1510,12 +1510,12 @@ paths:
               - picklist
         - name: after
           in: query
-          description: 'The cursor reference of the last entry for the previous page. Use the `end_cursor` value from the previous response to get the next page (`end_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `after=eyJzZXNzaW9uSWQiOjM4LCJmaWVsZElkIjo0MH0`. '
+          description: 'The cursor reference of the last entry for the previous page. Use the `end_cursor` value from the last response to get the next page (`end_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `after=eyJzZXNzaW9uSWQiOjM4LCJmaWVsZElkIjo0MH0`. '
           schema:
             type: string
         - name: before
           in: query
-          description: 'The cursor reference of the first entry for the next page. Use the `start_cursor` value from the previous response to get the previous page (`start_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `before=eyJzZXNzaW9uSWQiOjgsImZpZWxkSWQiOjMxfQ`.'
+          description: 'The cursor reference of the first entry for the next page. Use the `start_cursor` value from the last response to get the previous page (`start_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `before=eyJzZXNzaW9uSWQiOjgsImZpZWxkSWQiOjMxfQ`.'
           schema:
             type: string
     put:
--- a/reference/customers.v3.yml
+++ b/reference/customers.v3.yml
@@ -52,7 +52,7 @@ paths:
       parameters:
         - name: page
           in: query
-          description: Page number. `page` would be ignored when `before` or `after` is provided in the request. `page=1`
+          description: Page number (`page` will be ignored if `before` or `after` is provided in the request). For example `page=1`.
           schema:
             type: integer
         - in: query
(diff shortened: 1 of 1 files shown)


reference/customers.v3.yml
```diff
@@ -2233,7 +2253,55 @@ components:
                 items:
                   $ref: '#/components/schemas/customer_Full'
               meta:
-                $ref: '#/components/schemas/_metaCollection'
+                  $ref: '#/components/schemas/MetaOpen'
```
isaacchen-bc, line 2256: Update CustomerCollectionResponse meta to MetaOpen as we don't include pagination info in put post request

reference/customers.v3.yml
```diff
@@ -2233,7 +2253,55 @@ components:
...
+                    store_credit_amounts:
+                      - amount: 43.15
+                    accepts_product_review_abandoned_cart_emails: true
+                    channel_ids:
+                      - 1
+                    shopper_profile_id: "82511e54-4040-40fe-b742-2b25655f205b"
+                    segment_ids: 
+                      - "5bb733a9-5491-47b3-9451-9ae8d6a6bc6b"
+                meta: {}
+    GetCustomerCollectionResponse:
```
isaacchen-bc, line 2292: Introduce a new type for get endpoint

reference/customers.v3.yml
```diff
@@ -52,7 +52,7 @@ paths:
       parameters:
         - name: page
           in: query
-          description: Page number. `page=1`
+          description: Page number. `page` would be ignored when `before` or `after` is provided in the request. `page=1`
```
bc-akshatkaushik, line 55:
```suggestion
          description: Page number (`page` will be ignored if `before` or `after` is provided in the request). For example `page=1`.
```

reference/customers.v3.yml
```diff
@@ -208,9 +208,19 @@ paths:
               - 'last_name:desc'
               - 'date_modified:asc'
               - 'date_modified:desc'
+        - name: after
+          in: query
+          description: 'Use the `end_cursor` value from the previous response for the next page. For example `after=eyJpZCI6MjA0fQ`. `end_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`.'
```
bc-akshatkaushik, line 213:
```suggestion
          description: 'The cursor value `after` which page's results will be displayed.  Use the `end_cursor` value from the previous response for the next page (`end_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `after=eyJpZCI6MjA0fQ`. '
```

reference/customers.v3.yml
```diff
@@ -208,9 +208,19 @@ paths:
...
+          schema:
+            type: string
+        - name: before
+          in: query
+          description: 'Use the `start_cursor` value from the previous response for the previous page. For example `before=eyJpZCI6MjA1fQ`. `start_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`.'
```
bc-akshatkaushik, line 218:
```suggestion
          description: 'The cursor reference of the first entry of the next page. Use the `start_cursor` value from the previous response for the next page (`start_cursor` is only returned on the first page or when the request contains query parameter `before` or `after`). For example `before=eyJpZCI6MjA1fQ`.'
```

reference/customers.v3.yml
```diff
@@ -1452,7 +1462,7 @@ paths:
             default: application/json
         - name: page
           in: query
-          description: Page number. `page=1`
+          description: Page number. `page` would be ignored when `before` or `after` is provided in the request. `page=1`
```
bc-akshatkaushik, line 1465:
```suggestion
          description: Page number. `page` will be omitted when `before` or `after` is provided in the request. `page=1`
```
In the Response it will be absent so maybe `omitted` is more appropriate than `ignored`. `ignored` makes sense for the request where if a conflict is present it will be `ignored`

isaacchen-bc: This one is actually still the request 😄. The response one is called `pagination`

reference/customers.v3.yml
isaacchen-bc: > The cursor reference of the first entry of the next page. Use the `start_cursor` value from the previous response for the next page

we use before to get the previous page no?

reference/customers.v3.yml
```diff
@@ -52,7 +52,7 @@ paths:
       parameters:
         - name: page
           in: query
-          description: Page number. `page=1`
+          description: Page number (`page` will be ignored if `before` or `after` is provided in the request). For example `page=1`.
```
bc-traciporter, line 55:
```suggestion
          description: Page number (`page` will be ignored if you provide `before` or `after` in the request). For example `page=1`.
```

reference/customers.v3.yml
```diff
@@ -1452,7 +1462,7 @@ paths:
             default: application/json
         - name: page
           in: query
-          description: Page number. `page=1`
+          description: Page number (`page` will be ignored if `before` or `after` is provided in the request). For example `page=1`.
```
bc-traciporter, line 1465:
```suggestion
          description: Page number (`page` will be ignored if you provide `before` or `after` in the request). For example `page=1`.
```

bc-traciporter (commented):
Made some editorial comments to remove passive voice.

reference/customers.v3.yml
bc-akshatkaushik: Yeah so from the perspective of the current page where we will use this value, the page where we got the cursor from becomes the `next` page if that makes sense 😅 
We can also update to something like `first entry of the page following the one we are trying to fetch`

reference/customers.v3.yml
bc-akshatkaushik: Yeah I lost network access as I was trying to delete this comment 🤣

bc-akshatkaushik (approved):
Minor suggestion but otherwise LGTM 💯

reference/customers.v3.yml
isaacchen-bc: > The cursor reference of the first entry of the next page. 

The first part makes sense

> Use the start_cursor value from the previous response for the next page

But I think the second part should be using `start_cursor` as the value of `before` in order to get the previous page

snaderiBC (dismissed):
Thanks Isaac 👏 
Looking good 🚢

reference/customers.v3.yml
bc-akshatkaushik: Yeah true, maybe `current` page makes more sense here 🤔

bc-akshatkaushik (approved):
💯

bc-traciporter:
Hi @bc-akshatkaushik I plan to add this change to the release notes tomorrow. Here is my draft copy:

Cursor pagination is now available in the Customers V3 API for both the [Get All Customers](https://developer.bigcommerce.com/docs/rest-management/customers#get-all-customers) and [Get Customer Form Field Values](https://developer.bigcommerce.com/docs/rest-management/customers/form-field-values#get-customer-form-field-values) endpoints. This update simplifies navigation and supports more complex queries.

isaacchen-bc:
> Hi @bc-akshatkaushik I plan to add this change to the release notes tomorrow. Here is my draft copy:
> 
> Cursor pagination is now available in the Customers V3 API for both the [Get All Customers](https://developer.bigcommerce.com/docs/rest-management/customers#get-all-customers) and [Get Customer Form Field Values](https://developer.bigcommerce.com/docs/rest-management/customers/form-field-values#get-customer-form-field-values) endpoints. This update simplifies navigation and supports more complex queries.

Thanks for the draft @bc-traciporter! Just a bit of context for this. The reason why we introduce cursor pagination is that when users use it to navigate through the results, the performance would be drastically improved, especially when the store has a lot of customers or form field values.

bc-traciporter:
@bc-akshatkaushik Thanks for the feedback. I will change the text slightly.

Cursor pagination is now available in the Customers V3 API for both the [Get All Customers](https://developer.bigcommerce.com/docs/rest-management/customers#get-all-customers) and [Get Customer Form Field Values](https://developer.bigcommerce.com/docs/rest-management/customers/form-field-values#get-customer-form-field-values) endpoints. This update improves navigation performance and supports many customers or form field values.

PR #779, opened 2025-01-24
Added consents array to put and post request bodies

bc-traciporter:
#NO TICKET

## What changed?

Added the consents array to the PUT and POST request body schema and examples.
See slack conversation: https://bigcommerce.slack.com/archives/C57SQ401K/p1737710187469379

## Release notes draft

Enhancement

* 

## Anything else?

ping {names}

### DIFF
--- a/reference/subscribers.v3.yml
+++ b/reference/subscribers.v3.yml
@@ -382,10 +382,10 @@ components:
               type: string
             description: Shows what active subscriptions a shopper may have. If the consents array is empty, the user has unsubscribed or didn’t enable the newsletter subscription checkbox during checkout.
             enum:
-              - [marketing_newsletter]
-              - [abandoned_cart]
+              - ["marketing_newsletter"]
+              - ["abandoned_cart"]
             example: 
-              - [marketing_newsletter]
+              - ["marketing_newsletter"]
       x-internal: false
     subscriber_Base:
       title: subscriber_Base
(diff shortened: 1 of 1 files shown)


reference/subscribers.v3.yml
```diff
@@ -424,6 +424,14 @@ components:
           minimum: 1
           type: integer
           description: The channel ID where the subscriber was created.
+        consents:
+            type: array
+            items:
+              type: string
```
isaacchen-bc, line 430: Should we add an enum here? The value here can only be `marketing_newsletter`, `abandoned_cart`

bc-traciporter: @isaacchen-bc I added in the enum values for the consents array.

bc-traciporter:
@isaacchen-bc I finally structured the enum correctly!  Can you review my changes and approve?

reference/subscribers.v3.yml
```diff
@@ -380,9 +380,12 @@ components:
...
             items:
               type: string
-            description: Shows what active subscriptions a shopper may have. If the consents array is empty, the user has unsubscribed or didn’t enable the newsletter subscription checkbox during checkout.
+            description: Shows the active subscriptions a shopper may have. If the consents array is empty, the user has unsubscribed or didn’t enable the newsletter subscription checkbox during checkout. The array will contain consent types like 'marketing_newsletter' and 'abandoned_cart'.
+            enum:
+              - ["marketing_newsletter"]
+              - ["abandoned_cart"]
             example: 
-              - marketing_newsletter
+              - "marketing_newsletter"
```
isaacchen-bc, line 388: I think the enum needs to be under items as it's the case where the item is enum. Also the example needs to be an array. So I think it would look something like:
```
consents:
  type: array
  items:
    type: string
    enum:
      - marketing_newsletter
      - abandoned_cart
  description: ...
  example: ["marketing_newsletter"]
```

bc-traciporter: Thanks @isaacchen-bc, that worked!

isaacchen-bc (approved):
Thanks Traci!

PR #797, opened 2025-02-03
Ad `ttl` to shipping rates response body: [update] add missing ttl object to response body

bc-traciporter:
# NO TICKET

## What changed?

add missing ttl object to response body

See slack conversation: https://bigcommerce.slack.com/archives/C57SQ401K/p1738596845318689

* 

## Release notes draft

Bug Fix: Added missing optional field

* 

## Anything else?

ping {names}

### DIFF
--- a/reference/shipping_provider.yml
+++ b/reference/shipping_provider.yml
@@ -353,7 +353,7 @@ paths:
                 Without shipping rates:
                   value:
                     quote_id: example_quote
-                    ttl: 0
+                    ttl: 5
                     messages: []
                     carrier_quotes: []
       summary: Request shipping rates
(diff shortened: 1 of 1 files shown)


bc-0dp (approved):

reference/shipping_provider.yml
```diff
@@ -353,6 +353,7 @@ paths:
                 Without shipping rates:
                   value:
                     quote_id: example_quote
+                    ttl: 0
```
bc-0dp, line 356: We've found that a minimum of 5 is advisable here

reference/shipping_provider.yml
```diff
@@ -762,6 +763,8 @@ components:
         quote_id:
           type: string
           maxLength: 50
+        ttl:
+          type: number
```
bc-0dp, line 767: I think that this is an `integer` rather than a `number` as it will have no fractional component.

Furthermore, is it possible to add a description, for instance,

```
      ttl:
        type: number
        description: "Time-to-live (TTL) in seconds"
```

bc-0dp (approved):

PR #798, opened 2025-02-03
DEVDOCS-5007: [update] add redirect feature

bc-traciporter:
# [DEVDOCS-5007]

## What changed?

Added Redirect example under Get a Product.

* 

## Release notes draft

 Redirect parameter on the `route` node.

* 

## Anything else?

ping {names}

[DEVDOCS-5007]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/storefront/graphql/examples/products.mdx
```diff
@@ -87,6 +87,27 @@ Query all products by not including an argument for `products`.
 
 You can also query for featured products, related products, and more. See the [GraphQL Storefront Playground](https://developer.bigcommerce.com/graphql-storefront/playground) for full schema documentation.
 
+### Get a product by URL using a route node
+
+Query a list of products by matching the `Redirect` type.
```
bc-jz, line 92: You can only query a single resource (product in this case) at a time with the `route` node.

docs/storefront/graphql/examples/products.mdx
```diff
@@ -87,6 +87,27 @@ Query all products by not including an argument for `products`.
...
+### Get a product by URL using a route node
+
+Query a list of products by matching the `Redirect` type. 
+
+```graphql filename="Get a product with the redirect type" showLineNumbers copy
+
+query {
+  site {
+    route(path: "/laura-ashley-duck-egg-paint/113699-master/", redirect: [FOLLOW|IGNORE] = IGNORE) {
+      redirect # <--- new redirect node of a type Redirect
```
bc-jz, line 99: It does not seem relevant to include the `redirect` node in this response since the example is specific to getting a product. If you do include it though then it should be mentioned that `redirect` will only exist if the `path` supplied as input points at a redirection. Otherwise it will be `null`.

bc-traciporter: @bc-jz I am a little confused about this ticket. I am confused whether we should document the LookUpUrl capability, the redirect node, or the `route` and `node` nodes.  The title of the ticket says LookUpUrl to account for redirects. The GraphQL example in the ticket has a redirect node. Is this example querying a product or a redirect? Do we need a different example? Do we need two queries, one to query the redirect and then a second one to query a product. I really don't know how this works from just reading the tickets and looking at this example.

docs/storefront/graphql/examples/products.mdx
```diff
@@ -87,6 +87,27 @@ Query all products by not including an argument for `products`.
...
+      node {
+        id
+        ... on Product {
+          name
+        }
+      }
```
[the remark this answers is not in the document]
bc-traciporter: @bc-jz Do we need the routes.mdx file?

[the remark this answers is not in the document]
bc-jz: > @bc-jz Do we need the routes.mdx file?

I think having a specific `routes.mdx` would be good as it would allow describing the `route` node with depth that does not really apply to the example in this section. 

I'm thinking about organization of this information and trying to avoid making these examples overly complex by getting into details that are a bit beyond the needs of explaining how to "get a product by a url".

bc-traciporter:
@bc-jz I updated the new routes markdown file. Can you take a look and ensure it is technically accurate? I am wondering if this new article is enough or do you think I still need to add something to the [Routes](https://developer.bigcommerce.com/docs/storefront/headless/site-content#routes) section? Here is a preview of the text. You can see the graphQL examples in the PR. I added 3 examples (general query, query using IGNORE, and query using FOLLOW).



Screenshot 2025-02-12 at 3 46 18 PM

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
+# Routes with the GraphQL Storefront API
+
+Routes define specific URL paths for fetching or updating data. The BigCommerce GraphQL Storefront API allows you to retrieve routes for each storefront channel, making it easier to build headless storefront applications using BigCommerce's built-in content.
```
bc-jz, line 3: The `routes` node cannot perform updates. I would rewrite these sentences to something like:

```
The BigCommerce GraphQL Storefront API allows you to retrieve routes for each storefront channel, making it easier to build headless storefront applications using BigCommerce's built-in content. Use of the `route` node is based on a supplied URL. The response will be an entity that exists at that URL, a redirect to a new location, or nothing if the URL does not relate to an entity or redirect.
```

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
...
+* Product
+* Category
+* Brand
+* NormalPage
+* ContactPage
+* RawHtmlPage
+* Blog
+* BlogPost
+
+## Enhanced redirect support
```
bc-jz, line 16: I would just call this "Redirect support" since we have not previously documented this redirect behavior so adding the term "Enhanced" doesn't include a point of reference by which it is an enhancement.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+The new `LookUpUrl` capability now supports redirects, helping you determine the appropriate page to display. If a redirect exists for a given path, the `Redirect` node within the `route` query returns the redirect details. Otherwise, it returns `null`.
```
bc-jz, line 18: I don't think we should use the term "LookUpUrl". I don't think it has any meaning to a 3rd party.

I recommend rewriting this section as:

> The `route` node has support for 301 Redirects that have been added to the store. If the URL provided as input to the `route` references a redirect then the response will return a `Redirect` node which comes in two varieties: a manual redirect or a dynamic redirect.

> A `ManualRedirect` will return a fixed URL that is the destination of that redirect. A new `route` query will be needed to attempt to pull the content at that destination URL. This can lead to a chain of redirects if the destination URL is also referencing a redirect.

> A "dynamic redirect" returns an entity type and entity ID of an existing store resource. There are 5 types of dynamic redirects: `ProductRedirect`, `BrandRedirect`, `BlogPostRedirect`, `CategoryRedirect`, and `PageRedirect`. Once the entity type and ID are retrieve a new query can be initiated to pull that content from the appropriate resource node. However, it is possible to retrieve this content as part of the `route` query even if the URL provided is a dynamic redirect. We will detail further how this works below.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+## Redirect enum values
```
bc-jz, line 20: I suggest a title like:
> Redirect Behavior of Route Node

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+The `redirectBehavior` argument supports two values:
```
bc-jz, line 22: > It is possible to supply a `redirectBehavior` input argument which supports two values:

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
...
+               images {
+                 edges {
+                   node {
+                     url(width: 500, height: 500)
+                   }
+                 }
+               }
+             }
+             ... on Redirect {
+               # all redirect attributes
```
bc-jz, line 77: What does this comment mean?

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
...
+               toUrl
+             }
+           }
+         }
+       }
+     }
+```
+```json filename="GraphQL variables" showLineNumbers copy 
+    {
+      "urlPath": "/blog/"
```
bc-jz, line 87: Rather than provide an input URL of `/blog/` I think this example should have an input URL that references one of the entity types that we are defining in the query. Either a product, category, or brand. I might be confusing as to why the query did not define "Blog" as an response type yet that is what was returned.

I do think there is relevance to this type of example but we should clearly describe what is happening. I also don't think it should be the first example. The first example should be successful use of finding a desired entity at a URL.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
...
+          }
+        }
+      }
+   ```
+</Tab>
+</Tabs>
+
+### Query: Using `IGNORE` redirect behavior
+
+In this example query, the `node` value is "null" because the redirect is ignored.
```
bc-jz, line 111: I think it would be helpful to add a note here, ie:

> In this example query, the `node` value is "null" because the redirect is ignored. Note that the `redirectBehavior: IGNORE` input is default so could be excluded from the query and the same results would be received.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,204 @@
...
+    }
+  }
+}
+```
+</Tab>
+</Tabs>
+
+### Query: Using `FOLLOW` redirect behavior
+
+In this example, the `node` returns the target entity, in this case a product.
```
bc-jz, line 158: > In this example, the `node` returns the entity that is targeted by the dynamic redirect as part of the response because we chose to `FOLLOW` the redirect.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,269 @@
...
+* Brand
+* NormalPage
+* ContactPage
+* RawHtmlPage
+* Blog
+* BlogPost
+
+## Redirect support
+
+The `route` node supports 301 Redirects that you have been added to the store. If the URL provided as input to the `route`, references a redirect, then the response will return a `Redirect` node, which comes in two varieties: a manual redirect or a dynamic redirect.
```
bc-jz, line 20: > The `route` node supports 301 Redirects that you have been added to the store.

Looks like a typo...this should read as:
The `route` node supports 301 Redirects that have been added to the store.
or
The `route` node supports 301 Redirects that you have added to the store.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,269 @@
+# Routes with the GraphQL Storefront API
+
+The BigCommerce GraphQL Storefront API allows you to retrieve data at the destination of a route (URL) for each storefront channel using the `route` node. This option makes it easier to build headless storefront applications, removing the need to track entity ID and type when accessing data for entities such as Products, Categories, Pages, Blogs, and more.
+
+The only required input when using the `route` node is the destination URL. The response will be an entity at that URL, a redirect to a new location, or nothing if the URL does not relate to an entity or redirect.
```
bc-jz, line 5: Lets mention the input parameter here...I'm thinking a 2nd sentence in the existing paragraph like:

> The only required input when using the `route` node is the destination URL. Input is provided with the `path` param on the `route` node. The response will be an entity at that URL...

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,269 @@
...
+* Brand
+* NormalPage
+* ContactPage
+* RawHtmlPage
+* Blog
+* BlogPost
+
+## Redirect support
+
+The `route` node supports 301 Redirects that you have added to the store. If the URL provided as input to the `route`, references a redirect, then the response will return a `Redirect` node, which comes in two varieties: a manual redirect or a dynamic redirect.
```
bc-jz, line 20: > If the URL provided as input to the `route`, references a redirect, then the response will return a `Redirect` node, which comes in two varieties: a manual redirect or a dynamic redirect.

Improper commas in this sentence.

docs/storefront/headless/routes.mdx
bc-traciporter: Thank you @bc-jz  I changed line 5 as follows:

The destination URL is the only required input when using the route node. You will provide an input with the path parameter on the route node. The response will be an entity at that URL, a redirect to a new location, or nothing if the URL does not relate to an entity or redirect.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,269 @@
...
+
+A `ManualRedirect` returns a fixed destination URL. You will need a new `route` query to attempt to pull the content at that destination URL. If the destination URL itself is a redirect, this can create a chain of redirects requiring multiple queries.
+
+A "dynamic redirect" returns an entity type and entity ID of an existing store resource. There are five types of dynamic redirects: `ProductRedirect`, `BrandRedirect`, `BlogPostRedirect`, `CategoryRedirect`, and `PageRedirect`. Once the entity type and ID are retrieved, you can initiate a new query to pull that content from the appropriate resource node. With dynamic redirects, you can retrieve the target entity directly within the `route` query response, avoiding an additional request. We will detail how this works further below.
```
bc-jz, line 24: I'm not liking the use of "We" here since I don't think we are writing with this same perspective in other places. Maybe phrase this line:
> We will detail how this works further below.

as:
> This is detailed below.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,269 @@
...
+* **FOLLOW**: 
+
+  * The `node` node returns the target entity (e.g., category, product, blog, etc.) if the redirect is dynamic. 
+  * For static/manual redirects, the `node` node returns `null` (since there’s no associated entity), while the `redirect` node provides the redirect details.
+
+## Example queries
+
+### Query: Fetching a product
+
+This example retrieves details about a specific product based on a given URL path.
```
bc-jz, line 45: This example attempts to retrieve the entity at the path `/product_1`. In this case that URL is mapped to a dynamic redirect which is returned instead of the entity node. The redirect points at the destination entity which is a product with id `123` and has a URL of `/t-shirt/`.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,269 @@
...
+        }
+      }
+    }
+  }
+}
+   ```
+</Tab>
+</Tabs>
+
+### Query: Using `IGNORE` redirect behavior
```
bc-jz, line 174: This example is basically the same as your first one that is called "Query: Fetching a product".

I like this example and description better than that one and think you should probably just remove the first example altogether. Making "Query: Fetching route details" your first example makes more sense to me.

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,269 @@
...
+        "node": null
+      }
+    }
+  }
+}
+```
+</Tab>
+</Tabs>
+
+### Query: Using `FOLLOW` redirect behavior
```
bc-jz, line 221: 👍

bc-jz (approved):
There are a few more comments above for some changes but with those in place I think this document will be ready!

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,220 @@
...
+                 edges {
+                   node {
+                     url(width: 500, height: 500)
+                   }
+                 }
+               }
+             }
+             ... on Redirect {
+               toUrl
+             }
```
bc-jz, line 84: I think this is in the wrong spot. It looks like it is under the `node` block which it does not belong as it is not a type of object returned by the `node`. It should exist under a `redirect` node that sites at the same level as the `node` node. Or it could be removed altogether.

If you do place it under the `redirect` node then be sure to add it to the output as it will be `redirect: null` in this case.

bc-jz (approved):
That one change aside this is looking good! Nice work!

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,230 @@
+# Routes with the GraphQL Storefront API
```
zvuki, line 1:
```suggestion
# Storefront Routes with the GraphQL Storefront API
```

docs/storefront/headless/routes.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+The BigCommerce GraphQL Storefront API allows you to retrieve data at the destination of a route (URL) for each storefront channel using the `route` node. This option makes it easier to build headless storefront applications, removing the need to track entity ID and type when accessing data for entities such as Products, Categories, Pages, Blogs, and more.
```
zvuki, line 3:
```suggestion
The BigCommerce GraphQL Storefront API allows you to find the destination entity of a route (URL) for each storefront channel using the `route` node. This option makes it easier to build headless storefront applications, removing the need to track entity ID and type when accessing data for entities such as Products, Categories, Pages, Blogs, and more.
```

lukedsteiner (approved):
I have read through, and tested each example in my personal store. Everything works as described, and how we speak to it in the related courses for BigDev.

PR #799, opened 2025-02-04
DEVDOCS-6211 - update concurrent call limit (category trees)

bc-terra:
Added messaging about concurrent call rate limit to Get a Category Tree endpoint. This will be affected by an experiment in the next 90 days.

# [DEVDOCS-6211]

## What changed?

* The Get a Category Tree endpoint is getting a concurrent call limit to prevent overloading on the back-end.
* This endpoint is now documented.

## Release notes draft
* Updated Get a Category Tree endpoint to include mention of the default concurrent call limit.

## Anything else?

ping {names}

### DIFF
--- a/reference/catalog/category-trees_catalog.v3.yml
+++ b/reference/catalog/category-trees_catalog.v3.yml
@@ -402,7 +402,11 @@ paths:
   '/catalog/trees/{tree_id}/categories':
     get:
       summary: Get a category tree
-      description: Returns a category tree.
+      description: |-
+        Returns a category tree.
+
+        **Note:**
+        The default rate limit for this endpoint is 1 concurrent request.
       operationId: getCategoryTree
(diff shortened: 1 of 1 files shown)


olysenko (approved):

jyuricle:
We should put in on hold until all the [blockers](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-6211?focusedCommentId=1940720) are addressed

tomzweb:
@bc-terra This change broke several of our apps that depend on this route. We've contacted support and posted it on Slack, but have had no reply, so trying here as well.

We had to scramble to get a temporary fix out to get our apps working again, but perhaps you can advise on how we can manage this change when multiple users are using the app(s) at the same time?

It would have been nice to receive some notice of this update so we could have prepared.

PR #866, opened 2025-03-17
DEVDOCS-6312: [update] add cursor pagination

bc-traciporter:
# [DEVDOCS-6312]

## What changed?
Add cursor pagination to the following endpoints

- GET /v3/pricelists
- GET /v3/pricelists/{id}/records
- GET /v3/pricelists/assignments

## Release notes draft

* 

## Anything else?

ping {names}

[DEVDOCS-6312]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-6312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/price_lists.v3.yml
+++ b/reference/price_lists.v3.yml
@@ -221,6 +221,38 @@ paths:
                               The total number of pages in the collection.
                             example: 1
                         description: 'Data related to the response, including pagination and collection totals.'
+                      cursor_pagination:
+                        title: Pagination
+                        type: object
+                        properties:
+                          count:
+                            description: Total number of items in the collection response.
+                            type: integer
+                            format: int32
+                          per_page:
+                            description: The amount of items returned in the collection per page, controlled by the limit parameter.
+                            type: integer
+                            format: int32
+                          start_cursor:
+                            description: A string representing the starting point of the current page in the collection
+                            type: string
+                          end_cursor:
+                            description: A string representing the ending point of the current page in the collection.
+                            type: string
+                          links: 
+                            title: links
+                            description: Pagination links for the previous and next parts of the whole collection.
+                            type: object
+                            properties:
+                              previous:
+                                description: Link to the previous page returned in the response.
+                                type: string
+                              current:
+                                description: Link to the current page returned in the response.
+                                type: string
+                              next:
+                                description: Link to the next page returned in the response.
+                                type: string
                     description: 'Data related to the response, including pagination and collection totals.'
                 description: Get All PriceLists.
               example:
@@ -247,6 +279,12 @@ paths:
                     per_page: 50
                     current_page: 1
                     total_pages: 1
+                  cursor_pagination:
+                    count: 8
+                    per_page: 50
+                    start_cursor: "Nw=="
+                    end_cursor: "MTQ="
+                    links: {}
     post:
       tags:
         - Price Lists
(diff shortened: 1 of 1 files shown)


reference/price_lists.v3.yml
```diff
@@ -221,6 +221,38 @@ paths:
                               The total number of pages in the collection.
                             example: 1
                         description: 'Data related to the response, including pagination and collection totals.'
+                      cursor_pagination:
+                        title: Pagination
```
megdesko, line 225: Would we want to explicitly call this Cursor Pagination to distinguish from limit/ offset pagination around like 195?  What makes the most sense to end users?

reference/price_lists.v3.yml
```diff
@@ -221,6 +221,38 @@ paths:
...
+                        type: object
+                        properties:
+                          count:
+                            description: Total number of items in the collection response.
+                            type: integer
+                            format: int32
+                          per_page:
+                            description: The amount of items returned in the collection per page, controlled by the limit parameter.
```
megdesko, line 233:
```suggestion
                            description: The amount of items it is possible to return per page; this value is equal to the limit parameter, whether one is supplied or the default value is used.
```

megdesko: I may have gone into too much detail here. 😅

reference/price_lists.v3.yml
```diff
@@ -221,6 +221,38 @@ paths:
...
+                            type: integer
+                            format: int32
+                          start_cursor:
+                            description: A string representing the starting point of the current page in the collection
```
megdesko, line 237:
```suggestion
                            description: A string representing the starting point of the current page in the collection.
```
🍹  consistency since a lot of the other descriptions end in `.`

bc-traciporter:
Thanks for your feedback @megdesko. I made the editorial changes. Can you tell me what is the scenario for receiving either the pagination, cursor pagination or both in your response? I can write text but I need to structure it. Is it controlled by the limit parameter. Take a look at the [get coupon code](https://developer.bigcommerce.com/docs/rest-management/promotions/coupon-codes-bulk#get-coupon-codes) endpoint, do you want to do something similar?
Also do we need `before`, `after` parameters?

Here is the `meta` description for get coupon codes
```
Contains data about paginating the response via cursors. If no pagination details are specified, then both properties will be present. When a 'before' or 'after' cursor is provided, only the 'cursor_pagination' property will be present. When a 'page' parameter is provided, only the offset based 'pagination' property will be present.
```

megdesko:
Thanks @bc-traciporter - the text from coupon codes looks great!  And before/ after parameters are also needed (but optional, like page/ limit).

If only `limit` is passed we'll return both paginations but include that limit.

reference/price_lists.v3.yml
```diff
@@ -2032,13 +2184,6 @@ paths:
                   date_modified: '2022-09-18T13:18:15Z'
                   currency: eur
                   product_id: 185
-                meta:
```
megdesko, line 2035: Was this removed because we're looking at the singleton endpoint here?

megdesko (approved):
Thanks Traci for making these updates!

reference/price_lists.v3.yml
bc-traciporter: @megdesko that was a mistake. I added meta back in. 
When I tested this endpoint, the response looks as follows:

{
    "data": {
        "price_list_id": 8,
        "variant_id": 205,
        "price": 3.99,
        "sale_price": 0.1,
        "retail_price": 0.1,
        "map_price": 0.1,
        "calculated_price": 0.1,
        "date_created": "2025-03-17T15:47:48Z",
        "date_modified": "2025-03-17T15:52:25Z",
        "currency": "usd",
        "product_id": 141
    },
    "meta": {}
}

bc-terra (approved):

PR #871, opened 2025-03-19
DEVDOCS-6320: [update] Update a Web Analytics Provider request body

bc-traciporter:
# [DEVDOCS-6320]

## What changed?

Removed `channel_id` and `id` from the Update a Web Analytics Provider request body

## Release notes draft

* 

## Anything else?

ping {names}

### DIFF
--- a/reference/settings.v3.yml
+++ b/reference/settings.v3.yml
@@ -135,8 +135,6 @@ paths:
               description: ''
               type: object
               properties:
-                channel_id:
-                  type: number
                 code:
                   type: string
                   minLength: 1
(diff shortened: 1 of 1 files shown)


bc-rmalyavc (commented):
Looks like we are only removing `channel_id` from `body` here. We need to also add it as a `GET` parameter. This is the only way to update web analytics on channel level. We need to reflect it in the docs.

bc-rmalyavc:
> Looks like we are only removing `channel_id` from `body` here. We need to also add it as a `GET` parameter. This is the only way to update web analytics on channel level. We need to reflect it in the docs.

It turns out that we already have it in parameters. All good then

bc-rmalyavc (approved):
LGTM

PR #876, opened 2025-03-21
DEVDOCS-6321: [new] Data events API

bc-traciporter:
# [DEVDOCS-6321]

## What changed?
Create a new article for the Data Events API.

## Release notes draft

* 

## Anything else?

ping {names}

[DEVDOCS-6321]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-6321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/storefront/graphql/data-events.mdx
+++ b/docs/storefront/graphql/data-events.mdx
@@ -183,3 +183,9 @@ Navigate to the `redirectedCheckoutUrl` returned in the previous step, proceed t
 </Steps>
 
 By following these steps, you ensure that your headless storefront properly captures analytics events and sends relevant data to the BigCommerce analytics system.
+
+## Verification
+
+In the control panel, go to **Analytics** > **Marketing** and select the `Search` origin from the table to view the origins.
+
+Also, navigate to **Analytics** > **Purchase funnel** and observe the `Shopped` metric, which represents product views.
--- a/docs/storefront/graphql/data-events.mdx
+++ b/docs/storefront/graphql/data-events.mdx
@@ -6,7 +6,7 @@ This API is particularly useful for tracking analytical metrics, including visit
 
 ## Introduction
 
-A fundamental concept in BigCommerce analytics is the "visit" or "session". When sending events to BigCommerce you must include a `visitId` and `visitorId`, both formatted as UUID. You can generate these values using the [UUID Generator](https://www.uuidgenerator.net/version4).
+A fundamental concept in BigCommerce analytics is the "visit" or "session". When sending events to BigCommerce you must include a `visitId` and `visitorId`, both formatted as UUID. You can generate these values using the [UUID Generator](https://www.uuidgenerator.net/version4). The following steps, along with an example for enabling a headless storefront, are described below.
 
 ## Enabling a headless storefront
 
@@ -15,7 +15,7 @@ A fundamental concept in BigCommerce analytics is the "visit" or "session". When
 
 Execute the `VisitStartedEvent` mutation. You must provide the `visitId` and `visitorId` (in UUID format) in the `initiator` input. 
 
-Example request:
+Example:
 <Tabs items={['Request', 'Response']}>
   <Tab>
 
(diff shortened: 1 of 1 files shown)


docs/storefront/graphql/data-events.mdx
```diff
@@ -0,0 +1,191 @@
...
+### Complete the order
+
+Navigate to the `redirectedCheckoutUrl` returned in the previous step, proceed through checkout, and create an order.
+</Steps>
+
+By following these steps, you ensure that your headless storefront properly captures analytics events and sends relevant data to the BigCommerce analytics system.
+
+## Verification
+
+In the control panel, go to **Analytics** > **Marketing** and select the `Search` origin from the table to view the origins.
```
bc-rmalyavc, line 189: This step would work like this only if we specified `refererUrl` with value like `https://google.com` for the `VisitStarted` mutation. Otherwise the visit and order will be listed under `Direct`, not `Search`. I think it's fine to keep it simple and simply point to the `Direct` category in CP->marketing

docs/storefront/graphql/data-events.mdx
```diff
@@ -0,0 +1,189 @@
...
+        }
+      }
+    }
+  ```
+  </Tab>
+</Tabs>
+
+### Complete the order
+
+Navigate to the `redirectedCheckoutUrl` returned in the previous step, proceed through checkout, and create an order.
```
bc-rmalyavc, line 180: I'd like to add some basic explanation here also. Maybe not exactly but something like this
```suggestion
Navigate to the `redirectedCheckoutUrl` returned in the previous step, proceed through checkout, and create an order. At this point `Bigcommerce app` is sending some `Checkout` and `Order` related events to the Data pipeline. Navigating to the above `redirectedCheckoutUrl` makes sure that all further events will be tied to the same visit within analytics section.
```

docs/storefront/graphql/data-events.mdx
```diff
@@ -0,0 +1,189 @@
+# Data Events in the GraphQL Storefront API
+
+This article explains how to enable headless storefronts to utilize native analytics. The BigCommerce analytics system operates on data events. These are event objects that capture specific user actions such as adding a product to a cart, visiting a website, or placing an order. All of the BigCommerce analytics features rely on event data being transmitted to our data pipeline.
+
+This API is particularly useful for tracking analytical metrics, including visits and product views.
+
+## Introduction
+
+A fundamental concept in BigCommerce analytics is the "visit" or "session". When sending events to BigCommerce you must include a `visitId` and `visitorId`, both formatted as UUID. You can generate these values using the [UUID Generator](https://www.uuidgenerator.net/version4). The following steps, along with an example for enabling a headless storefront, are described below.
```
bc-rmalyavc, line 9: 🍸
I don't insist on it, but I thought, it would be good to explain these 2 values a bit more. Perhaps, you could phrase it better :) 
```suggestion
A fundamental concept in BigCommerce analytics is the visit details. When sending events to BigCommerce you must include a `visitId` and `visitorId`, both formatted as UUID. `visitId` represents session for the data pipeline. While `visitorId` represents a visitor/shopper. `visitId` is usually a short live cookie. In Stencil by default it's time to live is 30minutes. We suggest to use the same TTL on headless stores. `vistorId` on the contrary is assigned to a shopper and has TTL of 1 year on Stencil. You can generate these values using the [UUID Generator](https://www.uuidgenerator.net/version4). The following steps, along with an example for enabling a headless storefront, are described below.
```

bc-rmalyavc (dismissed):

docs/storefront/graphql/data-events.mdx
bc-traciporter: This is good information, yes I would love to include it.

bc-terra (approved):
Looks great!

PR #881, opened 2025-03-26
docs(orders): DEVDOCS-6330 [REST Storefront] [Orders] Add missing values for "include" query param

BC-SEven:
…

# [DEVDOCS-6330]

## What changed?

* Added missing values for include request param 

## Release notes draft

* 

## Anything else?

ping @bigcommerce/team-orders

### DIFF
--- a/reference/orders.sf.yml
+++ b/reference/orders.sf.yml
@@ -55,12 +55,22 @@ paths:
               type: string
               enum:
                 - lineItems
+                - lineItems.physicalItems
+                - lineItems.physicalItems.socialMedia
+                - lineItems.physicalItems.options
+                - lineItems.physicalItems.categories
+                - lineItems.digitalItems
+                - lineItems.digitalItems.socialMedia
+                - lineItems.digitalItems.options
+                - lineItems.digitalItems.categories
+                - lineItems.giftCertificates
                 - billingAddress
                 - coupons
                 - currency
                 - taxes
                 - payments
                 - consignments
+                - fees
       responses:
         '200':
           description: ''


reference/orders.sf.yml
```diff
@@ -55,12 +55,22 @@ paths:
               type: string
               enum:
                 - lineItems
+                - lineItems.physicalItems
+                - lineItems.physicalItems.socialMedia
+                - lineItems.physicalItems.options
+                - lineItems.physicalItems.categories
```
donald-nguyen-bc, line 61:
```suggestion
                - lineItems.physicalItems
                - lineItems.physicalItems.giftWrapping
                - lineItems.physicalItems.socialMedia
                - lineItems.physicalItems.options
                - lineItems.physicalItems.categories
```

donald-nguyen-bc, line 61: probably out of scope of this PR: should we also update the schema for the "ItemPhysical.categories" : https://github.com/bigcommerce/docs/blob/b874b6059d85db201ded82162e1dd9f6ed641270/reference/orders.sf.yml#L339-L344

reference/orders.sf.yml
```diff
@@ -55,12 +55,22 @@ paths:
               type: string
               enum:
                 - lineItems
+                - lineItems.physicalItems
+                - lineItems.physicalItems.socialMedia
+                - lineItems.physicalItems.options
```
donald-nguyen-bc, line 60: Seems like `options` attribute is not definied in the `ItemPhysical` 🤔

reference/orders.sf.yml
```diff
@@ -55,12 +55,22 @@ paths:
               type: string
               enum:
                 - lineItems
+                - lineItems.physicalItems
+                - lineItems.physicalItems.socialMedia
```
donald-nguyen-bc, line 59: `socialMedia` attribute also not defined in `ItemPhysical`

reference/orders.sf.yml
BC-SEven: I didn't include gift wrapping as that is always included by default. So adding the parameter wont have an effect

BC-SEven:
Ah I see we are still missing in the response body. Ill add those in as well.

PR #900, opened 2025-04-09
doc(orders): ORDERS-6930 update Refund API docs to include `tax_adjustment_amount`

matt-evangelidis:
# [ORDERS-6930]
Update `orders.v3.yml` to include a previously undocumented field `tax_adjustment_amount` for refunds and refund quotes

## What changed?

* Add new option including `tax_adjustment_amount` for POST request body for `/orders/{order_id}/payment_actions/refunds_quotes`, mutually exclusive with the existing `items` payload.
* Add new option including `tax_adjustment_amount` for POST request body for `/orders/{order_id}/payment_actions/refunds`, mutually exclusive with the existing `items` payload.
* Extend description of `items` in the GET/POST response body for `/orders/{order_id}/payment_actions/refunds` to indicate that the `items` array will be empty if `tax_adjustment_amount` was used to create the refund.

image

image

image

## Release notes draft

* 

## Anything else?

ping @bigcommerce/team-orders

[ORDERS-6930]: https://bigcommercecloud.atlassian.net/browse/ORDERS-6930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/orders.v3.yml
+++ b/reference/orders.v3.yml
@@ -2295,17 +2295,12 @@ components:
         $ref: '#/components/schemas/RefundQuote_Post'
       x-internal: false
     RefundQuote_Post:
-      type: object
       title: RefundQuote_Post
       description: Request body for refund quotes.
       x-internal: false
-      properties:
-        items:
-          type: array
-          items:
-            $ref: '#/components/schemas/ItemsRefund'
-      required:
-        - items
+      anyOf:
+        - $ref: '#/components/schemas/RefundQuote_ItemsRefund'
+        - $ref: '#/components/schemas/RefundQuote_TaxAdjustmentAmount'
     RefundQuote_Full:
       type: object
       title: RefundQuote_Full
@@ -2391,10 +2386,15 @@ components:
           items:
             $ref: '#/components/schemas/RefundMethod'
     RefundRequest_Post:
-      type: object
       description: Request body for refund requests.
       title: RefundRequest_Post
       x-internal: false
+      anyOf:
+          - $ref: '#/components/schemas/RefundRequest_Post_Items'
+          - $ref: '#/components/schemas/RefundRequest_Post_TaxAdjustmentAmount'
+    RefundRequest_Post_Items:
+      title: Items Refund
+      type: object
       properties:
         items:
           type: array
@@ -2409,6 +2409,16 @@ components:
       required:
         - items
         - payments
+    RefundRequest_Post_TaxAdjustmentAmount:
+      title: Tax Adjustment Refund
+      type: object
+      properties:
+        tax_adjustment_amount:
+          $ref: '#/components/schemas/TaxAdjustmentAmount'
+        merchant_calculated_override:
+          $ref: '#/components/schemas/MerchantOverride'
+      required:
+        - tax_adjustment_amount
     RefundID_Get:
       type: object
       x-examples:
@@ -2504,6 +2514,7 @@ components:
                     description: |
                       Message indicates why the payment was declined.
             items:
+              description: Array of items refunded. In cases when `tax_refund_adjustment` was used to create the refund, this array will be empty.
               type: array
               items:
                 type: object
@@ -2760,6 +2771,37 @@ components:
         - total_amount
         - total_tax
       x-internal: false
+    TaxAdjustmentAmount:
+      type: number
+      format: float
+      title: Tax Adjustment Amount
+      example: 1.99
+      description: Amount to be used when tax may have been overcharged for an order, such as when the value for a partial refund is overridden. This amount should be equal to the calculated overcharged value, or should be used with `merchant_calculated_override` to override the value. If not, this will result in a `422` error.
+      x-internal: false
+    RefundQuote_ItemsRefund:
+      title: Items Refund
+      type: object
+      x-internal: false
+      properties:
+        items:
+          type: array
+          items:
+            $ref: '#/components/schemas/ItemsRefund'
+        merchant_calculated_override:
+          $ref: '#/components/schemas/MerchantOverride'
+      required:
+        - items
+    RefundQuote_TaxAdjustmentAmount:
+      title: Tax Adjustment Refund
+      type: object
+      x-internal: false
+      properties:
+        tax_adjustment_amount:
+          $ref: '#/components/schemas/TaxAdjustmentAmount'
+        merchant_calculated_override:
+          $ref: '#/components/schemas/MerchantOverride'
+      required:
+        - tax_adjustment_amount
     Refund:
       type: object
       title: Refund
@@ -2793,8 +2835,7 @@ components:
           description: Whether refund amount and tax are provided explicitly by merchant override.
         items:
           type: array
-          description: Array of items refunded.
-          minItems: 1
+          description: Array of items refunded. In cases when `tax_refund_adjustment` was used to create the refund, this array will be empty.
           items:
             $ref: '#/components/schemas/RefundItem'
         payments:


BC-SEven (dismissed):

YevenLourance (approved):

reference/orders.v3.yml
```diff
@@ -2760,6 +2769,33 @@ components:
...
+            $ref: '#/components/schemas/ItemsRefund'
+      required:
+        - items
+    RefundQuote_TaxAdjustmentAmount:
+      type: object
+      x-internal: false
+      properties:
+        tax_adjustment_amount:
+          $ref: '#/components/schemas/TaxAdjustmentAmount'
+        merchant_calculated_override:
```
donald-nguyen-bc, line 2795: do we have this field in "refund quote" request ?

matt-evangelidis: Shockingly, yes.

https://github.com/bigcommerce/bigcommerce/blob/master/app/Order/Application/Api/V1/PaymentActions/Refund/Validator/RefundQuoteValidator.php#L15

image

donald-nguyen-bc: oh, I see. Probably we will need to add this field to `RefundQuote_ItemsRefund` as well, similar to `RefundRequest_Post_Items`

matt-evangelidis: Already exists on `RefundRequest_Post_items`, but I'll add it to `RefundQuote_ItemsRefund`.

reference/orders.v3.yml
```diff
@@ -2760,6 +2769,35 @@ components:
         - total_amount
         - total_tax
       x-internal: false
+    TaxAdjustmentAmount:
+      type: number
+      format: float
+      title: Tax Adjustment Amount
+      example: 1.99
+      description: Amount to be used when tax may have been overcharged for an order, such as when the value for a partial refund is overridden. This amount should be equal to the calculated overcharged value, or should be used with `merchant_calculated_override` to override the value. If not, this will result in a `422` error.
```
donald-nguyen-bc, line 2777: I'm curious if it is only used for "overcharged" ? what about "undercharged" scenario ?

matt-evangelidis: If you do an "undercharged" scenario, which I characterise as the following:
- Refunding a line item, and then overriding the refund value to be less than the original total and tax. E.g. a line item of 119.95 + 11.995 (10%) in tax for a total of 131.945 is refunded for 118

You end up with some unintuitive behaviour from the UI and the API

image

`refundable-order` response
```json
{
    "data": {
        "id": 130,
        "customer": {
            "first_name": "sample",
            "last_name": "sample",
            "phone": "+10400000000",
            "email": "test@address.com",
            "stats": {
                "order_count": 1,
                "refund_count": 1
            }
        },
        "billing_address": {
            "first_name": "sample",
            "last_name": "sample",
            "company": "sample",
            "street_1": "11305 4 Points Dr Building 2",
            "street_2": "Suite 100",
            "city": "Austin",
            "zip": "78726",
            "country": "United States",
            "state": "Texas"
        },
        "date_created": "2025-04-11T01:16:54+00:00",
        "date_created_formatted": "Apr 11th 2025",
        "custom_status": "Partially Refunded",
        "is_tax_included": false,
        "shipping_amount": 0,
        "handling_amount": 0,
        "total_inc_tax_amount": 186.4,
        "total_tax_amount": 16.95,
        "total_refunded_amount": 118,
        "total_refundable_amount": 68.4,
        "order_level_refund_amount": 0,
        "payment_provider": "Test payment provider",
        "payment_method": "Credit Card",
        "store_credit_amount": 0,
        "gift_certificate": null,
        "origin": {
            "name": "Storefront",
            "is_internal": true
        },
        "addresses": [
            {
                "first_name": "sample",
                "last_name": "sample",
                "company": "sample",
                "street_1": "11305 4 Points Dr Building 2",
                "street_2": "Suite 100",
                "city": "Austin",
                "zip": "78726",
                "country": "United States",
                "state": "Texas",
                "shipping_item": {
                    "id": 31,
                    "base_amount": 0,
                    "refunded_amount": 0,
                    "total_discount_amount": 0,
                    "applied_discounts": []
                },
                "handling_item": null,
                "product_items": [
                    {
                        "id": 37,
                        "product_id": 37,
                        "name": "[Sample] Chemex Coffeemaker 3 Cup",
                        "sku": "CC3C",
                        "base_amount": 49.5,
                        "total_discount_amount": 0,
                        "quantity": 1,
                        "refunded_quantity": 0,
                        "wrapping": null,
                        "applied_discounts": [],
                        "product_options": [],
                        "configurable_fields": [],
                        "type": "PHYSICAL",
                        "is_digital": false
                    },
                    {
                        "id": 36,
                        "product_id": 36,
                        "name": "[Sample] Tiered Wire Basket",
                        "sku": "TWB",
                        "base_amount": 119.95,
                        "total_discount_amount": 0,
                        "quantity": 1,
                        "refunded_quantity": 1,
                        "wrapping": null,
                        "applied_discounts": [],
                        "product_options": [],
                        "configurable_fields": [],
                        "type": "PHYSICAL",
                        "is_digital": false
                    }
                ]
            }
        ],
        "refundable_tax_adjustment_amount": 12,
        "tax_provider": "Basic Tax",
        "fees": [],
        "currency": {
            "currency_symbol": "$",
            "currency_symbol_position": "left",
            "decimal_places": 2
        }
    },
    "meta": {}
}
```

image

POST `refund_quotes`

```json
{
    "items": [],
    "tax_adjustment_amount": 12
}
```

```json
{
    "data": {
        "total_refund_amount": 13.95,
        "total_refund_tax_amount": 12,
        "rounding": 0,
        "adjustment": 0,
        "is_tax_included": false,
        "order_level_refund_amount": 0,
        "refund_methods": [
            [
                {
                    "provider_id": "bigpaypay",
                    "provider_description": "Test payment provider",
                    "amount": 13.95,
                    "offline": true,
                    "offline_provider": false,
                    "offline_reason": "This provider does not support refund."
                }
            ],
            [
                {
                    "provider_id": "custom",
                    "provider_description": "Custom",
                    "amount": 13.95,
                    "offline": true,
                    "offline_provider": true,
                    "offline_reason": "This is an offline payment provider."
                }
            ]
        ]
    },
    "meta": {}
}
```

donald-nguyen-bc: sorry I'm not sure I understand 100% of your example above, does it mean we can use this field `tax_adjustment_amount` for the undercharged tax as well ?

matt-evangelidis: Sorry for not illustrating it better. My main point here is that it's very easy to end up with behaviour that's very hard to follow. As far as I'm aware, when doing a line-item refund, we can't really undercharge for tax, because we're limited in the total amount we can refund (equal to `total_inc_tax` for the line item), even when overriding.

In this example, I refunded a line item that was $131.945 (inc tax), but I overrode the value to be $118, which is less than the original exc_tax amount of $119.95.

When I come back to do another refund, the system states there may be an overcharge of $12. This comes from the original refund state (without the override), where the price and tax looked like this:

image

That $12 is the tax component missing, but not the whole missing value (131.95 - 118 = 13.95)

So despite seeing $12 show up as the missing amount and minimum we can refund, we instead see $13.95 come up as the new value when navigating.


https://github.com/user-attachments/assets/1242bc48-3f35-44cd-9b3b-5ac86fc58edc

donald-nguyen-bc: thanks for the explanation Matthew 🙇 , all good

donald-nguyen-bc (approved):

jpajar (dismissed):

bc-traciporter:
Hi @matt-evangelidis and @donald-nguyen-bc I reviewed this PR. I was a little confused with the output. I noticed that the organization of the current doc looks different than what I see using the spec editor. Which is correct?

[Current doc](https://developer.bigcommerce.com/docs/rest-management/transactions/payment-actions#create-a-refund)
Screenshot 2025-04-14 at 8 42 11 AM

Spec editor (https://developer.bigcommerce.com/spec-editor)
Screenshot 2025-04-14 at 8 42 21 AM

matt-evangelidis:
> Hi @matt-evangelidis and @donald-nguyen-bc I reviewed this PR. I was a little confused with the output. I noticed that the organization of the current doc looks different than what I see using the spec editor. Which is correct?
> 
> [Current doc](https://developer.bigcommerce.com/docs/rest-management/transactions/payment-actions#create-a-refund) Screenshot 2025-04-14 at 8 42 11 AM
> 
> Spec editor (https://developer.bigcommerce.com/spec-editor) Screenshot 2025-04-14 at 8 42 21 AM

Hi, @bc-traciporter, thanks for calling this out (I didn't know https://developer.bigcommerce.com/spec-editor existed prior to this 😅). The above appearance definitely isn't intended.

It seems to be a result of me changing the schema to use `oneOf`, like the following

```yaml
    RefundRequest_Post:
      description: Request body for refund requests.
      title: RefundRequest_Post
      x-internal: false
      oneOf:
          - $ref: '#/components/schemas/RefundRequest_Post_Items'
          - $ref: '#/components/schemas/RefundRequest_Post_TaxAdjustmentAmount'
```

This seems to work and display as intended elsewhere, but in our own solution, it doesn't seem to like or support `oneOf` as a top-level field, even though that would be semantically correct in this case. We can get around this by using `anyOf` to display it correctly, so I'll update the PR to use that instead.

image

bc-traciporter:
HI @matt-evangelidis Thank you for looking into this, the schema is tricky.  The Create a Refund Quote endpoint looks good. However, the Create a Refund endpoint looks like it still needs to be updated. The any Of: items displays above the items array. All other payment actions endpoints look good.

matt-evangelidis:
Hi @bc-traciporter, thanks for catching that, just missed adding the title to the other bit of the schema. Updated to include for the `Create a Refund` endpoint

image

bc-traciporter (approved):
Approved

bc-traciporter:
Thank you @matt-evangelidis for your work, the changes look good. Are you able to merge the PR or do you need us to do it?

PR #946, opened 2025-05-14
chore(orders): ORDERS-6254 make the quantity bound item first so the …

bc-vincent-zhao:
…refund adjustment feature is more visible and easier to find

# [DEVDOCS-]

## What changed?

* Just swap the ordering of different refund item type and put QuantityBoundItem first because that contains refund adjustment feature

Right now, the `adjustment` field is not very visible on the doc because `Amount bound item` is shown by default, only upon click on the `Quantity bound item` it shows the `adjustments` field.  It may cause friction when users searching for the field 

Screenshot 2025-05-14 at 10 48 44 am

This simple change makes the `QuantityBoundItem` the first default so `adjustments` feature is more visible/searchable.

## Release notes draft

ping @bigcommerce/team-orders @bigcommerce/dev-docs

### DIFF
--- a/reference/orders.v3.yml
+++ b/reference/orders.v3.yml
@@ -2576,8 +2576,8 @@ components:
       title: ItemsRefund
       x-internal: false
       oneOf:
-        - $ref: '#/components/schemas/AmountBoundItem'
         - $ref: '#/components/schemas/QuantityBoundItem'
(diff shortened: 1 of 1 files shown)


donald-nguyen-bc (approved):
👍 the change makes sense

PR #968, opened 2025-06-04
docs(promotions): PROMO-869 split promotion models into Draft, Patch …

BC-SEven:
…and Saved as each state has different expectations on what props are required

## What changed?
Refactor promotion models to separate them into "Draft", "Patch" and "Saved" models to be used depending on context.
Here we can correct expectations for the SavedPromotions so id is always available. And for patch payloads where no properties are required.

## Release notes draft
No behaviour changes here

## Anything else?

ping @bigcommerce/team-promotions


reference/promotions.v3.yml
```diff
@@ -1144,9 +1286,9 @@ components:
                 minItems: 1
                 type: string
                 example: "0439538491"
-    CouponCode:
+    DraftCouponCode:
```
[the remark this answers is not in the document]
BC-SEven: Ah yea. Actually no need for this Draft model. Its not used

reference/promotions.v3.yml
```diff
@@ -1158,11 +1297,6 @@ components:
           description: 'A unique code that can be used to manually apply a discount. Only letters, numbers, white spaces, underscores and hyphens are allowed.'
           example: TEST-COUPON-CODE
           maxLength: 50
-        current_uses:
```
jpajar, line 1161: I noticed this was removed, are we no longer tracking `current_uses` for `CouponCode`?

BC-SEven: Ah good catch. No this shouldn't be removed. 😅

reference/promotions.v3.yml
```diff
@@ -382,16 +389,67 @@ components:
...
+          properties:
+            codes:
+              $ref: '#/components/schemas/CouponCode'
+            coupon_overrides_automatic_when_offering_higher_discounts:
+              type: boolean
+              example: false
+              default: false
+              description: |-
+                This field only has effect when `can_be_used_with_other_promotions` is `false`:
+                - When the property is set to 'true', the coupon will override the applied automatic promotions if it provides a greater discount.
```
bc-terra, line 407:
```suggestion
                - When the property is set to `true`, the coupon will override the applied automatic promotions if it provides a greater discount.
```

reference/promotions.v3.yml
```diff
@@ -382,16 +389,67 @@ components:
...
+                - When the property is set to 'false', the coupon will not be applied if automatic promotions are already applied.
+                Trying to set the value of this field to 'true' when `can_be_used_with_other_promotions` is `true` will yield a 422 error response.
```
bc-terra, line 409:
```suggestion
                - When the property is set to `false`, the coupon will not be applied if automatic promotions are already applied.
                Trying to set the value of this field to `true` when `can_be_used_with_other_promotions` is `true` will yield a 422 error response.
```

reference/promotions.v3.yml
```diff
@@ -382,16 +389,67 @@ components:
...
+              $ref: '#/components/schemas/CouponCode'
+            coupon_overrides_automatic_when_offering_higher_discounts:
+              type: boolean
+              example: false
+              default: false
+              description: |-
+                This field only has effect when `can_be_used_with_other_promotions` is `false`:
+                - When the property is set to 'true', the coupon will override the applied automatic promotions if it provides a greater discount.
+                - When the property is set to 'false', the coupon will not be applied if automatic promotions are already applied.
+                Trying to set the value of this field when `can_be_used_with_other_promotions` is `true` will yield a 422 error response.
```
bc-terra, line 427:
```suggestion
                - When the property is set to `true`, the coupon will override the applied automatic promotions if it provides a greater discount.
                - When the property is set to `false`, the coupon will not be applied if automatic promotions are already applied.
                Trying to set the value of this field when `can_be_used_with_other_promotions` is `true` will yield a 422 error response.
```

reference/promotions.v3.yml
```diff
@@ -402,11 +460,90 @@ components:
                 - When the property is set to "false", the coupon will not be applied if automatic promotions are already applied.
```
bc-terra, line 460:
```suggestion
                - When the property is set to `false`, the coupon will not be applied if automatic promotions are already applied.
```

PR #979, opened 2025-06-16
DEVDOCS-6364 - Add brands to translations

bc-terra:
- Updated language to be inclusive of both brands and categories
- Added callouts related to content.
- Adjusted format to use tabs more consistently

# [DEVDOCS-6364]

## What changed?
* Translations GraphQL queries and mutations now support translations for brands

## Release notes draft
* We've added Brands to our growing list of resources whose translations can be managed via GraphQL

## Anything else?

ping { @bigcommerce/dev-docs-team  }

### DIFF
--- a/docs/store-operations/translations/categories.mdx
+++ b/docs/store-operations/translations/categories.mdx
@@ -1,57 +1,71 @@
-# Translations for Categories (Beta)
+# Translations for Categories and Brands (Beta)
 
 <Callout type="info">
 The Translations Admin GraphQL API is currently available on Catalyst storefronts only.
 </Callout>
 
+The BigCommerce GraphQL API provides queries and mutations for retrieving and managing translations in stores operating within a multi-language context. This allows for a seamless experience for customers across multiple locales.
 
-The categories translatable fields are:
+The current implementation supports managing translations for categories and brands, handled using analogous queries and mutations separated by the `resourceType` filter and `resourceID`. The table below provides the details for each of these fields as they would be included in queries and mutations:
 
-* Name
-* Description
-* Page Title
-* Meta Keywords
-* Meta Description
-* Search Keywords
+| Resource | resourceType | resourceId |
+| --- | --- | --- |
+| Category | `CATEGORIES` | `bc/store/category/{{categoryId}}` |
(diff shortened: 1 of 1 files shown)


docs/store-operations/translations/categories.mdx
```diff
@@ -1,57 +1,71 @@
-# Translations for Categories (Beta)
+# Translations for Categories and Brands (Beta)
```
kristinapototska, line 1: can we make brands being a separate section not coupled with categories

kristinapototska: Let's structure it as:
Categories
Brands
Locations

in the side nav. 
Each will have it's own page with API schema, examples, list of translatable fields etc

kristinapototska (changes requested):

docs/store-operations/translations/categories.mdx
bc-yehor-kravchenko: Yes, it's better to create `brands.mdx`

bc-terra: I can see why you might want to approach it that way, but structurally the two processes are nearly synonymous. By separating it into two pages, we add extra steps and remove clarity.

Specifically, if we look at the structure of a brands request, there are only two differences between it and a categories request, both of which are non-structural. Addressing them on the same page makes it clear how they are related and how they are distinct.

I'd like to wait @bc-kpab to get back from vacation to weigh in on this one, as documentation structure is also part of our general style guide.

kristinapototska: @bc-terra the general idea behind this API is the the structure is all the same and there are just different entity types. Separating those in sections gives developers more clarity on choosing the entity they need instead of scrolling through one large page.
E.g. if I only need to add translations for brands and categories and products are already translated, or my app only interacts with categories and I don't need any extra data. 

Granular structure in the docs allow developers to access the data they need without scrolling large pages trying to get to the entity they are looking for. 

We expect this granular structure for all entities we will be adding to this API.

kristinapototska: This is also consistent with other GQL Admin API docs structure, where we granularly separate each content area in a separate section to allow easy and quick navigation.

PR #985, opened 2025-06-20
docs(search): ES-5682 - Clarify filtering by attributes behavior

jfugalde:
# [ES-5682 - Clarify filtering by attributes behavior]

## What changed?

* Clarification on how filtering by disabled filters can potentially malfunction or change without further notice

## Release notes draft

*  This is just a quick: when merchants rely on filtering by disabled filters, there’s no guarantee that behavior will remain consistent. Since these filters are not marked as enabled, any changes to the filtering engine, facets logic, or performance optimizations can cause these filters to silently break, malfunction, or disappear from results — without any prior notice.

In short, depending on disabled filters is relying on undefined behavior — and that makes the implementation fragile by design.

## Anything else?

ping {@bc-terra}

### DIFF
--- a/docs/storefront/headless/products/faceted-textual-search.mdx
+++ b/docs/storefront/headless/products/faceted-textual-search.mdx
@@ -232,7 +232,9 @@ You can also search and filter to include only products that have a specific pro
 
 <Callout type="info">
 - Product attributes are based on product variants and custom fields.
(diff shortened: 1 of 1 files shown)


bc-terra (approved):
Looks great. If this is a net new change, please be sure to attach a CSS ticket to the ES ticket this cites.

PR #1028, opened 2025-07-31
DEVDOCS-6241 - Content Refresh - "Address" (S2S)

bc-Vince:
Added depth and considerations to endpoint descriptions, parameters, and fields. Reorganized schemas to reduce field overlap.

# [DEVDOCS-6241](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-6241)

## What changed?

* Descriptions in the introduction and for each endpoint provide more information
* Rquest and response bodies reflect current intended format
* Parameters and fields include helpful considerations
* Each endpoint lists error responses with explanations of why the error might occur
* Schemas contain fewer redundant lines

## Release notes draft

* As part of our ongoing content refresh for B2B Edition developer documentation, we've fully refreshed the Server-to-Server Addresses documentation. This includes:
    * New, thorough descriptive copy
    * Request body examples with relevant fields
    * Inclusion of previously-missing fields for all endpoints

## Anything else?

ping @bc-terra

### DIFF
--- a/docs/b2b-edition/specs/api-v3/address/address.yaml
+++ b/docs/b2b-edition/specs/api-v3/address/address.yaml
@@ -1,21 +1,52 @@
 openapi: '3.0.0'
 info:
-  title: Address
+  title: Addresses
   version: v3
-  description: BigCommerce B2B Edition Address
+  description: |-
+    B2B Edition Company users can store several billing and shipping addresses to their Company account. This allows buyers in the Company to quickly supply address information to quotes, invoices, and orders.
+
+    The v3 Server to Server Addresses API allows you to manage billing and shipping addresses on behalf of your Company accounts. Use these endpoints to retrieve addresses, edit their details, and create new ones.
   contact:
     name: BigCommerce
     url: 'https://www.bigcommerce.com'
     email: support@bigcommerce.com
 servers:
   - url: 'https://api-b2b.bigcommerce.com/api/v3/io'
+tags:
+  - name: Addresses
 paths:
   /countries:
     parameters: []
     get:
-      summary: Get Countries
+      summary: Get a Country
       tags:
-        - Address
+        - Addresses
+      operationId: get-addresses-countries-countryName-code
+      description: |-
+        Retrieves country information based on the country name or two-letter code entered in the parameters. The response can be used to provide a valid country name or code in Company addresses.
+
+        Use the `searchType` parameter to specify whether you are searching by a country's name or its two-letter ISO code, and then add the appropriate value for the country to the `q` parameter. Use `https://store-{store_hash}.mybigcommerce.com/internalapi/v1/store/countries` to return a list of valid country names and codes for the `q` value.
+      parameters:
+        - schema:
+            type: integer
+            enum:
+              - 0
+              - 1
+            default: 0
+            example: 0
+          in: query
+          name: searchType
+          description: Determines whether you are searching by a country's name (`0`) or its two-letter ISO code (`1`).
+          required: true
+        - schema:
+            type: string
+            example: US (the searchType should be 1);  United States (the searchType should be 0)
+          in: query
+          name: q
+          description: The country's name or its two-letter ISO code, depending on the specified `searchType`. Use `https://store-{store_hash}.mybigcommerce.com/internalapi/v1/store/countries` to return a list of valid country names and codes.
+          required: true
+      security:
+        - authToken: []
       responses:
         '200':
           description: OK
@@ -23,19 +54,11 @@ paths:
           content:
             application/json:
               schema:
-                description: ''
                 type: object
-                x-examples:
-                  example-1:
-                    code: 200
-                    data:
-                      - countryCode: US
-                        countryName: United States
-                    meta:
-                      message: SUCCESS
                 properties:
                   code:
                     type: number
+                    default: 200
                   data:
                     required:
                       - countryCode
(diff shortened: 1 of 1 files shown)


docs/b2b-edition/specs/api-v3/address/address.yaml
```diff
@@ -213,54 +106,84 @@ paths:
...
+      tags:
+        - Addresses
       operationId: get-addresses-states-stateName-code
-      description: Get a state's info from its code/name.
+      description: |-
+        Retrieves state or province information based on the state name or two-letter code entered in the parameters. The response can be used to provide a valid state name or code in Company addresses.
+
+        Use the `country` parameter to specify the country associated with the queried state. If left blank, the endpoint uses the United States by default.
+        
+        Note that only the countries listed below will return state or province values. When [creating](#create-a-company-address) or [updating](#update-an-address) addresses in these countries, you must include state information.
```
CNanninga, line 128: Just a thought here:  Do we have "hard-coded" info, so to speak, in core BC docs about details like countries that have fixed state/province lists?  How frequently does that info change?

My first instinct is that it's unnecessary to go into detail on info like this that come become outdated, and that it would be sufficient that only some countries return values or require them when creating/updating addresses.  But that depends a lot on what we do elsewhere in the docs and how frequently the info changes.

docs/b2b-edition/specs/api-v3/address/address.yaml
```diff
@@ -271,372 +194,411 @@ paths:
...
+                    default: 200
                   data:
-                    type: array
-                    uniqueItems: true
-                    minItems: 1
-                    items:
-                      $ref: https://raw.githubusercontent.com/bigcommerce/docs/main/docs/b2b-edition/models/address/address.yaml
+                    required:
+                      - stateCode
+                      - stateyName
```
CNanninga, line 201: Looks like a typo here: "stateyName"

CNanninga (changes requested):
This is marked "Request changes" only because of possible typo.

For the comment about listing countries that accept states, feel free to disregard it.

docs/b2b-edition/specs/api-v3/address/address.yaml
bc-Vince: @CNanninga  When writing this, I was unable to find existing documentation in the Dev Center. My understanding is that this doesn ot change often, but we can circumvent any potential changes by removing the list and simply recommending the use of the v1 endpoint to confirm countries with valid/required states. What do you think about that?

CNanninga: Summary from DM convo:

If we have an endpoint that will return the information, I think that would be better than the hard-coded list.

However, it came up that the endpoint in question, which is also referenced elsewhere in this PR, is `https://store-{store_hash}.mybigcommerce.com/internalapi/v1/store/countries`, which is unfamiliar to me and which I'm not certain is documented anywhere.  I think this information needs to be vetted with the broader P&E team to make sure it's a valid practice we're recommending.  (And if so, I would recommend a follow-up ticket to make sure this endpoint is documented appropriately if it's not already.)

docs/b2b-edition/specs/api-v3/address/address.yaml
```diff
@@ -1,195 +1,88 @@
...
-                    data: {}
-                    meta:
-                      message: Resource not found
+        - Addresses
       operationId: get-addresses-countries-countryName-code
-      description: Get a country's info from its code/name.
+      description: |-
+        Retrieves country information based on the country name or two-letter code entered in the parameters. The response can be used to provide a valid country name or code in Company addresses.
+
+        Use the `searchType` parameter to specify whether you are searching by a country's name or its two-letter ISO code, and then add the appropriate value for the country to the `q` parameter. Use `https://store-{store_hash}.mybigcommerce.com/internalapi/v1/store/countries` to return a list of valid country names and codes for the `q` value.
```
bc-Vince, line 28:
```suggestion
        Use the `searchType` parameter to specify whether you are searching by a country's name or its two-letter [ISO code](https://www.iso.org/obp/ui/#search), and then add the appropriate value for the country to the `q` parameter. This endpoint only accepts country name values in English, but you can set the `searchBy` paramteter to `1` to retrieve country information with the standard ISO code instead.
```

docs/b2b-edition/specs/api-v3/address/address.yaml
```diff
@@ -213,54 +106,84 @@ paths:
...
+          description: Determines whether you are searching by a state or province's name (`0`) or its two-letter ISO code (`1`). This parameter also defines the value type for the `country` parameter.
           required: true
         - schema:
             type: string
             example: TX (the searchType should be 1);  Texas (the searchType should be 0)
           in: query
           name: q
-          description: Input state name/code
+          description: |-
+            The state's name or its two-letter ISO code, depending on the specified `searchType`. Use `https://store-{store_hash}.mybigcommerce.com/internalapi/v1/store/countries` to return a list of valid country names and codes.
```
bc-Vince, line 176:
```suggestion
            The state's name or its two-letter ISO code, depending on the specified `searchType`.
```

docs/b2b-edition/specs/api-v3/address/address.yaml
```diff
@@ -213,54 +106,84 @@ paths:
...
             default: US / United States
           in: query
           name: country
-          description: 'Country code or name,it should correspond to the search type you have chosen.'
-  /addresses:
-    get:
-      summary: Get Addresses
-      tags:
-        - Address
+          description: The country's name or its two-letter ISO code, depending on the specified `searchType`. Use `https://store-{store_hash}.mybigcommerce.com/internalapi/v1/store/countries` to return a list of valid country names and codes.
```
bc-Vince, line 186:
```suggestion
          description: The country's name or its two-letter ISO code, depending on the specified `searchType`.
```

CNanninga (approved):

bc-NickV (approved):
approved.

PR #1038, opened 2025-08-07
Mark order_taxes.tax_rate_id as deprecated, document sales_tax_id

theromulans:
# [TAX-2463]

## What changed?
* Mark `order_taxes.tax_rate_id` as deprecated.
* Document `order_taxes.sales_tax_id` (this has actually actually been in the response for a while now).

## Release notes draft
- To support a wider range of use cases, the `order_taxes.tax_rate_id` (int) is now deprecated and will be removed in the near future. Please update to use the `order_taxes.sales_tax_id` (string) field.

[TAX-2463]: https://bigcommercecloud.atlassian.net/browse/TAX-2463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/orders.v2.oas2.yml
+++ b/reference/orders.v2.oas2.yml
@@ -2816,6 +2816,7 @@ components:
                   order_id: 138
                   order_address_id: 39
                   tax_rate_id: 1
+                  sales_tax_id: 'TaxId123'
                   tax_class_id: 0
                   name: Tax
                   class: Default Tax Class
@@ -2828,6 +2829,7 @@ components:
                   order_id: 138
                   order_address_id: 40
                   tax_rate_id: 1
+                  sales_tax_id: 'TaxId123'
                   tax_class_id: 0
                   name: Tax
                   class: Default Tax Class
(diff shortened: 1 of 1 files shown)


benpratt77 (dismissed):

declankirk (approved):

reference/orders.v2.oas2.yml
```diff
@@ -3373,6 +3375,11 @@ components:
           description: The unique numeric identifier of the tax rate.
```
donald-nguyen-bc, line 3375: 🍹 probably we should mention about using `sales_tax_id` instead of this `tax_rate_id` field

```suggestion
          description: The unique numeric identifier of the tax rate. Note: this field has been deprecated, please use sales_tax_id instead
```

theromulans: Good idea @donald-nguyen-bc, thanks!

PR #1080, opened 2025-09-14
DEVDOCS-6496 - Clarify date format for modifiers in documentation

bc-terra:
Updated documentation for date modifier to specify ISO-8601 ATOM format.

# [DEVDOCS-6496]

## What changed?
* When sending a PUT or POST request to create/update product modifier options, the full ISO-8601 ATOM format is required.
* Currently, if the wrong format is provided, such as “2005-08-15”, validation should fail, but it doesn’t.

## Release notes draft
* We've updated the validation workflow for creating product modifiers to require ISO-8601 ATOM format, which includes date, time, and timezone. Failing to provide dates in ISO-8601 ATOM format will now result in a validation error.

## Anything else?

ping { @bigcommerce/dev-docs-team }


chris-nowicki (approved):
LGTM 🚀

